PR #18, opened 2021-05-25
feat: DT-466 Support `dev.email` trace tag

richardlarocque:
[DT-466](https://outreach-io.atlassian.net/browse/DT-466)

Updates the `trace.yaml` config file to support a new section,
`GlobalTags`.  The idea here is that `GlobalTags` will allow us to
specify in configs certain tags we would like to add to all traces.  The
`MarshalLog` implementation for the `GlobalTags` config struct will
handle converions to standardized tag names.

We start with just one attribute: `DevEmail`/`dev.email`.  This will be
injected into pods via bootstrap so that apps running in the local dev
environment will be tagged with the developer's email, making the
Honeycomb `dev` dataset much easier to navigate.

This is a forward-port of
https://github.com/getoutreach/go-outreach/pull/165.

### DIFF
--- a/pkg/trace/config.go
+++ b/pkg/trace/config.go
@@ -6,7 +6,18 @@ import (
 
 // tracing config goes into trace.yaml
 type Config struct {
-	Honeycomb `yaml:"Honeycomb"`
+	Honeycomb  `yaml:"Honeycomb"`
+	GlobalTags `yaml:"GlobalTags,omitempty"`
+}
+
+type GlobalTags struct {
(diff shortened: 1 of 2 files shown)


jaredallard (approved):
LGTM. Thanks!

richardlarocque:
@jaredallard Is it safe to ignore the bootstrapabot failure?

jaredallard:
> @jaredallard Is it safe to ignore the bootstrapabot failure?

Yep! This is a separate issue I need to work out with the OSS migration

PR #22, opened 2021-06-07
feat: DT-466 Make some tags appear on all spans

richardlarocque:
[DT-466](https://outreach-io.atlassian.net/browse/DT-466)

Updates the go-outreach trace logic with a `presendHook` to add certain
service-level tags to every outgoing span.  For now, these include the
`app.Info` tags and the `dev.email` tag.  I expect more will be added in
the future.

These extra tags help to enrich the "leaf" spans of the trace, making
large "GROUP BY"-style queries more practical.  We use this all the time
in the flagship, for example summing DB time by `rails.controller`.
This gets us started towards doing the same for Go services.

Honeycomb charges us by the even (ie. span) not by the size of those
spans, so these extra tags are effectively free.


richardlarocque:
Forward-porting from https://github.com/getoutreach/go-outreach/pull/169/commits.

jaredallard (approved):
LGTM. Thanks! ❤️

PR #24, opened 2021-06-17
feat: use existing ssh agent if present

zaquestion:

If merged intent would be to pull this into bootstrap so that it can be
run without prompting for a passpharase on every run.

Will also check the agent's existing keys when adding a new key so no
prompting is needed if the agent already contains the key.

**Notes for your reviewer**:

Tested this locally by replacing the version of `gobox` that bootstrap uses with this version. I then built `bootstrap` like so
```
go build  -ldflags "-X github.com/getoutreach/go-outreach/v2/pkg/app.Version=v6.13.0" -o bin/bootstrap cmd/bootstrap/bootstrap.go
```
This way when executing the binary it would try to clone a real version of itself. Was able to skip the passphrase check and clone the repo using the existing agent.

### DIFF
--- a/pkg/sshhelper/ssh_auth.go
+++ b/pkg/sshhelper/ssh_auth.go
@@ -4,6 +4,7 @@ package sshhelper
 import (
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"path/filepath"
 
@@ -22,7 +23,15 @@ import (
 // GetSSHAgent creates a new ssh-agent, originally it would return one only
 // if it was needed, but it's probably better to always use our own.
 func GetSSHAgent() agent.Agent {
-	return agent.NewKeyring()
+	a := agent.NewKeyring()
+	if addr, ok := os.LookupEnv("SSH_AUTH_SOCK"); ok {
+		agentConn, err := (&net.Dialer{}).Dial("unix", addr)
+		if err == nil {
+			a = agent.NewClient(agentConn)
+		}
+	}
+
+	return a
 }
 
 // GetPasswordInput retrieves input that doesn't echo back to the user as they type it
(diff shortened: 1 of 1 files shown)


pkg/sshhelper/ssh_auth.go
```diff
@@ -61,6 +70,25 @@ func AddKeyToAgent(keyPath string, a agent.Agent, log logrus.FieldLogger) error
 	if err != nil {
 		return errors.Wrap(err, "failed to read private key")
 	}
+	pubByts, err := ioutil.ReadFile(keyPath + ".pub")
```
jaredallard, line 73: Can we derive the public key from the private key? I'm weary about forcing a `.pub` to exist, this library is used in a lot of different "environments" at the moment and this could break things

zaquestion: Yeah, I wasn't fond of this either, actually looks like that should be doable now that I'm taking a second look at it. I'll fiddle with that.

zaquestion:
I did notice in testing that `bootstrap` will fallback to the ssh agent if configuring the one it pulls out of gobox fails. This still seemed like a solid improvement to the lib, but it's feasible we could also update bootstrap to skip this whole `configureAuth` step when an ssh agent it present.

pkg/sshhelper/ssh_auth.go
zaquestion: Hmm, I'm not sure if deriving the pubkey is possible without decrypting the private key (providing the passphrase) itself, which is what we're trying to skip. It seems like it should be tho...?

zaquestion: From some quick googling https://blog.tinned-software.net/generate-public-ssh-key-from-private-ssh-key/ seems to imply you'd need the passphrase, and this is what the available libraries imply as well.

jaredallard: Hmmmm, yes that wouldn't be possible I think. I'd say we'd have to do that here, taking a dependency on `.pub` is to concerning to me at the moment

pkg/sshhelper/ssh_auth.go
```diff
@@ -61,6 +70,25 @@ func AddKeyToAgent(keyPath string, a agent.Agent, log logrus.FieldLogger) error
...
+	keys, err := a.List()
+	if err != nil {
+		return errors.Wrap(err, "failed to list existing agent keys")
+	}
+	for _, k := range keys {
+		if ssh.FingerprintSHA256(pubKey) == ssh.FingerprintSHA256(k) {
+			// key already in agent
+			return nil
+		}
+	}
```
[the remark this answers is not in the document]
jaredallard: This works!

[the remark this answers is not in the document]
zaquestion: Updated :D

pkg/sshhelper/ssh_auth.go
zaquestion: Updated by making this not required in the function

jaredallard (approved):
LGTM. Thanks!!

PR #29, opened 2021-07-09
feat: Separate Latency Metrics [DT-479]

george-e-shaw-iv:
* Separate latency metrics into HTTP, gRPC, and Outbound.

**JIRA ID**: [DT-479]

**Notes for your reviewer**:
* Bootstrap will be updated after this is merged.

[DT-479]: https://outreach-io.atlassian.net/browse/DT-479

### DIFF
--- a/pkg/trace/call.go
+++ b/pkg/trace/call.go
@@ -2,6 +2,7 @@ package trace
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/getoutreach/gobox/pkg/app"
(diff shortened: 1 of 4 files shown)


pkg/trace/call.go
```diff
@@ -93,12 +186,27 @@ func EndCall(ctx context.Context) {
...
+
+	if info.kind == metrics.CallKindExternal {
+		info.ReportOutboundLatency()
+	} else {
+		switch info.name { //nolint:exhaustive //Why: we only report latency metrics in this case on HTTP/gRPC call types.
+		case CallTypeHTTP:
+			info.ReportHTTPLatency()
+		case CallTypeGRPC:
+			info.ReportGRPCLatency()
+		}
```
vk-outreach, line 198: You could probably add a `callType.reportLatency(info)` and move the switch over to there.

vk-outreach (approved):
LGTM

jaredallard (approved):
LGTM. Glad we're doing this

pkg/trace/call.go
george-e-shaw-iv: nice idea, i'll do that

PR #31, opened 2021-07-10
Add error metadata for use with grpc 

zpatrick:
This adds a new error type that embeds metadata, which must be of type `map[string]string` in order to be used with the [errdetails.ErrorInfo](https://pkg.go.dev/google.golang.org/genproto@v0.0.0-20210708141623-e76da96a951f/googleapis/rpc/errdetails#ErrorInfo) type.

This is meant to support https://github.com/getoutreach/services/pull/134

### DIFF
--- a/pkg/orerr/orerr.go
+++ b/pkg/orerr/orerr.go
@@ -86,6 +86,15 @@ func WithRetry() ErrOption {
 	}
 }
 
+// WithMeta attaches the provided grpc metadata to the error.
+//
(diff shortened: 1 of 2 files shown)


vk-outreach (commented):
LGTM but until we force a replace directive mapping go-outreach to gobox, we should probably update go-outreach as well.

george-e-shaw-iv (approved):

PR #35, opened 2021-07-16
feat: add FromString

zpatrick:


jaredallard (commented):
late approval, but here I am :D

PR #37, opened 2021-07-20
feat: add box library

jaredallard:
**What this PR does / why we need it**: This PR adds the box configuration library from the devenv into `gobox`. 

**JIRA ID**: [DTSS-493]

**Notes for your reviewer**:

[DTSS-493]: https://outreach-io.atlassian.net/browse/DTSS-493


george-e-shaw-iv (changes requested):
sry to do this to you but all exposed symbols + the package directive in the file that shares the same name as the package should have documentation

lintroller is coming soon

george-e-shaw-iv (approved):

PR #48, opened 2021-08-30
feat: Force Sample Rate for Tracing [DTSS-721]

george-e-shaw-iv:
* Allow traces to force their sample rates.

**JIRA ID**: [DTSS-721]

**Notes for your reviewer**:

[DTSS-721]: https://outreach-io.atlassian.net/browse/DTSS-721

### DIFF
--- a/pkg/trace/force_sampler.go
+++ b/pkg/trace/force_sampler.go
@@ -48,8 +48,8 @@ func forceTracing(ctx context.Context) context.Context {
 	return ctx
 }
 
-// sampleAt sets the sample rate for a given trace explicitly.  
-// The rate specifies how many samples were looked at for each 
+// sampleAt sets the sample rate for a given trace explicitly.
+// The rate specifies how many samples were looked at for each
 // accepted sample.  That is, the actual sampling  = 1/rate.
 func sampleAt(ctx context.Context, rate uint) context.Context {
 	if t := trace.GetTraceFromContext(ctx); t != nil {
--- a/pkg/trace/force_sampler.go
+++ b/pkg/trace/force_sampler.go
@@ -10,11 +10,13 @@ import (
 type samplerHook func(map[string]interface{}) (bool, int)
 
 const (
-	fieldTraceID    = "trace.trace_id"
-	fieldForceTrace = "force_trace"
+	fieldTraceID     = "trace.trace_id"
+	fieldForceTrace  = "force_trace"
+	fieldSampleTrace = "sample_trace"
 )
 
 // forceSampler allows force sample rate to 100% when trace context contains field force_trace
+// and sample at a rate of 1/<given rate> when context contains field sample_trace.
 func forceSampler(sampleRate uint) samplerHook {
 	sampler, err := sample.NewDeterministicSampler(sampleRate)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


bhogatemadhura (approved):

pkg/trace/trace.go
```diff
@@ -198,6 +198,12 @@ func ForceTracing(ctx context.Context) context.Context {
 	return forceTracing(ctx)
 }
 
+// ForceSampleRate will force a desired sample rate for the given trace and all children
+// of said trace. The sample rate in practice will be 1/<rate>.
+func ForceSampleRate(ctx context.Context, rate uint) context.Context {
```
jaredallard, line 203: This is slightly confusing, can we provide a `ExampleSampleRate`?

jaredallard (approved):
Let's give it a shot :tada:

pkg/trace/trace.go
george-e-shaw-iv: How you'd invoke this in my opinion isn't confusing which is the value of `Example*` functions, but I will add more to that comment. I'll add an explicit example that maths out a sample rate

pkg/trace/force_sampler.go
```diff
@@ -38,3 +47,12 @@ func forceTracing(ctx context.Context) context.Context {
 	}
 	return ctx
 }
+
+// sampleAt sets the sample rate for a given trace explicitly. The rate given will be
+// parsed as 1/rate to actually determine the sample rate.
```
vk-outreach, line 52: nit: the wording here (probably copied from my own from elsewhere) can be improved a little.

```suggestion
// sampleAt sets the sample rate for a given trace explicitly.  
// The rate specifies how many samples were looked at for each 
// accepted sample.  That is, the actual sampling  = 1/rate.
```

pkg/trace/force_sampler_test.go
```diff
@@ -66,3 +66,48 @@ func (suite) TestForceSamplerWhenNotForced(t *testing.T) {
 		}
 	})
 }
+
+func testSampleAtSampler(ctx context.Context, sampleRate uint, handler func(context.Context, *tracetest.TraceLog)) {
+	defer app.SetName(app.Info().Name)
+	app.SetName("gobox")
+
+	trlog := tracetest.NewTraceLogWithOptions(tracetest.Options{
+		SamplePercent: 1,
```
vk-outreach, line 75:
```suggestion
		SamplePercent: 0,
```

vk-outreach (approved):
LGTM

I think this should work

pkg/trace/force_sampler.go
```diff
@@ -38,3 +47,12 @@ func forceTracing(ctx context.Context) context.Context {
...
+func sampleAt(ctx context.Context, rate uint) context.Context {
+	if t := trace.GetTraceFromContext(ctx); t != nil {
+		t.AddField(fieldSampleTrace, rate)
+	}
+	return ctx
+}
```
vk-outreach, line 58: FWIW, I'd not have done it this way but this is ok.

I'd have done it as a log marshaler:

```golang

type TraceSampleRate  uint
func (t TraceSampleRate) MarshalLog(addField func(k string, v interface{})) {
    addField(fieldSampleTrace, rate)
}
```

I believe if we did this way, you can just do:

```golang
    ctx = trace.StartCall(ctx, "foo", log.TraceSampleRate(1000))
```

But we don't expect to use this a whole lot, so it doesn't matter a whole lot.

PR #56, opened 2021-09-15
feat(tester): add tester package

vk-outreach:
**What this PR does / why we need it**:   Adds a tester package compatible with testing.TB
**JIRA ID**: [DTSS-0]
**Notes for your reviewer**:

This PR just migrates synthetics/pkg/tester over here as it is a rather small & basic package.    The synthetics repo can now be archived once this lands.

I updated the document comments a bit to be more accurate and renamed LogResultsResults which was incorrectly stuttering + added the tests and also migrated the logger to gobox rather than go-outreach.

### DIFF
--- a/pkg/tester/tester.go
+++ b/pkg/tester/tester.go
@@ -1,0 +1,356 @@
+// Package tester implements a test runner compatible with testing.T
+//
+// Usage:
+//
+//     t := tester.New()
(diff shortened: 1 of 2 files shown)


vk-outreach:
FYI: @getoutreach/cs-aaaa

bhogatemadhura (approved):

PR #57, opened 2021-09-16
feat: VI-3634 add AZURE_BENTO env

guoweis-outreach:
We're trying to deploy mint in azure. The main blocker is that azure doesn't have such bento/namespace convention. This causes all sort of hack, like transcription-frontdoor (in namespace `rolling`) needs to lie about the namespace it's in such that it can talk to mint in mint--dev-westus2 namespace, and had to create an empty namespace (like `frontdoor--dev-westus2`) just to host a serviceaccount, such that mint can query k8s api to get the annotation of that sa account.

This relaxes the requirement of bento naming convention and solve this problem.

**JIRA ID**: [VI-3634]

**Notes for your reviewer**:

[VI-3634]: https://outreach-io.atlassian.net/browse/VI-3634

### DIFF
--- a/pkg/app/info.go
+++ b/pkg/app/info.go
@@ -70,9 +70,13 @@ func info() *Data {
 	// well enough for now.
 	serviceID := fmt.Sprintf("%s@outreach.cloud", appName)
 
-	parts := strings.Split(namespace, "--")
-	if len(parts) == 2 {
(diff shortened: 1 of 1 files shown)


jaredallard:
FYI this is an open-source repository

PR #74, opened 2021-10-25
feat: log context

clevelittlefield-outreach:
Adds a new context to log to propagate core fields using the central library.  This is similar to `trace.AddInfo` but has some key differences:

- MarshalLog is invoked immediately instead of dynamically
- The results are constrained to a finite set of value types

This is intended to be used sparingly, but since this is external OSS, we cannot effectively limit what fields we can supply here.  Perhaps we could set the limit on the number of fields supported, like 5 for example?

This is separate from trace.AddInfo, because we might have values there we do not want to share with logging.   This would also preserve this being an opt in, if you are not interested in these fields, if you don't create a Log context the execution will be the same as before.

### DIFF
--- a/pkg/log/log_context.go
+++ b/pkg/log/log_context.go
@@ -40,33 +40,28 @@ func filterAllowList(info F) F {
 var infoKey = "54be8dc9-91ac-4f77-b90a-70e1ffd74566" //random guid
 
 // Creates a new Value context to store fields that should be attached to all logs
-func NewContext(ctx context.Context) context.Context {
-	if infoKeyVal := ctx.Value(infoKey); infoKeyVal != nil {
-		return ctx
+// MarshalLog is invoked immediately on all args to reduce risk of hard to debug issues
+// and arbitrary code running during logging.
+func NewContext(ctx context.Context, args ...Marshaler) context.Context {
+	returnCtx := ctx
+	infoKeyVal := ctx.Value(infoKey)
+	if infoKeyVal == nil {
+		infoKeyVal = &F{}
+		// we use a guid string to avoid versioning issues, should not have collisions
+		returnCtx = context.WithValue(ctx, infoKey, infoKeyVal) //nolint:revive, staticcheck
 	}
 
-	// we use a guid string to avoid versioning issues, should not have collisions
-	return context.WithValue(ctx, infoKey, &F{}) //nolint:revive, staticcheck
+	logInfo := infoKeyVal.(fieldsSet)
+	temp := F{}
+	many := Many(args)
+	many.MarshalLog(temp.Set)
+	temp = filterAllowList(temp)
+	temp.MarshalLog(logInfo.Set)
+	return returnCtx
 }
 
 type fieldsSet interface {
 	Set(field string, value interface{})
-}
-
-// Add arguments to all logs. Return true if this is a log info context after args are added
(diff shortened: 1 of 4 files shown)


pkg/log/log_context.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		result[k] = v
+	}
+
+	return result
+}
+
+// nolint:gochecknoglobals
+var infoKey = "54be8dc9-91ac-4f77-b90a-70e1ffd74566" //random guid
```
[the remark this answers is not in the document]
kadukf-outreach: I meant switching from fixed string to private type alias + key as the instance of the alias - so that the context key is not accessible from outside. (also public example: https://github.com/open-telemetry/opentelemetry-go/blob/19294aab4cd8b52afc34b77403e7447109c9b858/trace/context.go#L21)

If I know the guid I can access the context data w/o issues + it is too long.

kadukf-outreach:
@clevelittlefield-outreach I added 1 comment but looking forward to it on master :) so that I can use it

pkg/log/log_context.go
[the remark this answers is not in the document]
clevelittlefield-outreach: Ahh got it.  This looks good, implementing.

kadukf-outreach (approved):

vk-outreach:
The change itself lgtm but will let dev-tooling folks do the stamp.

Can we prepare the bootstrap changes to set AllowContextFields to the tracing IDs + authn Identities?  If folks use this before we change bootstrap, the bootstrap changes to call AllowContextFields would become breaking changes which I'd like to avoid.

pkg/log/log_context.go
```diff
@@ -0,0 +1,75 @@
...
+		// we use a guid string to avoid versioning issues, should not have collisions
+		returnCtx = context.WithValue(ctx, currentLogContextKey, infoKeyVal) //nolint:revive, staticcheck
+	}
+
+	logInfo := infoKeyVal.(fieldsSet)
+	temp := F{}
+	many := Many(args)
+	many.MarshalLog(temp.Set)
+	temp = filterAllowList(temp)
+	temp.MarshalLog(logInfo.Set)
```
kadukf-outreach, line 60: @clevelittlefield-outreach when we nest contexts, shouldnt we create a copy of already existing infokeyval, merge then new data and then set it on the context? otherwise when the exit the inner. context then we still have data from the inner context.

pls, could you add the nested usecase and see what is on the parent and nested context?

clevelittlefield-outreach: Good point, adjusted.  I would hope we are not nesting these contexts too much but this change handles that situation.

pkg/log/log_context.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	return result
+}
+
+type logContextKeyType int
+
+const currentLogContextKey logContextKeyType = iota
+
+// Creates a new Value context to store fields that should be attached to all logs
```
jaredallard, line 43: `// NewContext ...` format

pkg/log/log_context.go
```diff
@@ -0,0 +1,67 @@
+package log
+
+import (
+	"context"
+)
+
+var allowList map[string]bool
+
+func AllowContextFields(fields ...string) {
```
jaredallard, line 9: This should have a comment in the form of `// AllowContextFields ...` as it's exported

PR #75, opened 2021-10-28
fix: Stop update on context cancellation

plisy:
Updater ignored SIGINT before.

**JIRA ID**: [DTSS-557]

## Notes for your reviewers
After this, the download is cancelled, but extraction will still finish if it started. To prevent that we'd have to wrap the reader in something context aware.

[DTSS-557]: https://outreach-io.atlassian.net/browse/DTSS-557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/updater/github.go
+++ b/pkg/updater/github.go
@@ -28,6 +28,7 @@ import (
 var (
 	ErrNoNewRelease = errors.New("no new release")
 	ErrNoAsset      = errors.New("no asset found")
+	ErrMissingFile  = errors.New("file missing in archive")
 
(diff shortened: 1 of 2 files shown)


jaredallard (approved):
LGTM, but we should probably test this by using `replace` in the `devenv` or something and building an "older" copy with `make APP_VERSION=v0.0.0` to get the update logic to trigger `./bin/devenv --force-update-check`

PR #78, opened 2021-11-04
feat: abstract out majority of bootstrap CLI code

jaredallard:

This PR pulls most of the CLI-specific code we have out of bootstrap and instead, putting it inside of gobox. This makes the code much easier to maintain / test and is less duplicated.

Example PR on the devenv: https://github.com/getoutreach/devenv/pull/130

**JIRA ID**: [DTSS-0]

### DIFF
--- a/pkg/cli/bootstrap.go
+++ b/pkg/cli/bootstrap.go
@@ -28,18 +28,16 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-const (
-	// UpdateExitCode is the exit code returned when an update ocurred
-	UpdateExitCode = 5
-)
+// UpdateExitCode is the exit code returned when an update ocurred
+const UpdateExitCode = 5
 
 // overrideConfigLoaders fakes certain parts of the config that usually get pulled
 // in via mechanisms that don't make sense to use in CLIs.
-func overrideConfigLoaders(honeycombApiKey, dataset string) {
+func overrideConfigLoaders(honeycombAPIKey, dataset string) {
 	var fallbackSecretLookup func(context.Context, string) ([]byte, error)
 	fallbackSecretLookup = secrets.SetDevLookup(func(ctx context.Context, key string) ([]byte, error) {
 		if key == "APIKey" {
-			return []byte(honeycombApiKey), nil
+			return []byte(honeycombAPIKey), nil
 		}
 
 		return fallbackSecretLookup(ctx, key)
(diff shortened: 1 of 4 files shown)


pkg/cli/bootstrap.go
```diff
@@ -0,0 +1,255 @@
...
+	"github.com/getoutreach/gobox/pkg/log"
+	"github.com/getoutreach/gobox/pkg/secrets"
+	"github.com/getoutreach/gobox/pkg/trace"
+	"github.com/getoutreach/gobox/pkg/updater"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
+	"gopkg.in/yaml.v2"
+)
+
+const (
```
george-e-shaw-iv, line 31: const block needs a comment for lintroller platinum, even tho this isn't platinum we might as well get with the program

pkg/cli/bootstrap.go
```diff
@@ -0,0 +1,255 @@
...
+	var fallbackSecretLookup func(context.Context, string) ([]byte, error)
+	fallbackSecretLookup = secrets.SetDevLookup(func(ctx context.Context, key string) ([]byte, error) {
+		if key == "APIKey" {
+			return []byte(honeycombApiKey), nil
+		}
+
+		return fallbackSecretLookup(ctx, key)
+	})
+
+	log.SetOutput(io.Discard)
```
george-e-shaw-iv, line 48: is this just to make sure we're using logrus over gobox log

pkg/cli/bootstrap.go
```diff
@@ -0,0 +1,255 @@
...
+
+// urfaveRegisterShutdownHandler registers a signal notifier that translates various term
+// signals into context cancel
+func urfaveRegisterShutdownHandler(cancel context.CancelFunc) {
+	// handle ^C gracefully
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
+	go func() {
+		<-c
+		cancel()
```
george-e-shaw-iv, line 93:
```suggestion
		<-c
		signal.Reset()
		cancel()
```

pkg/cli/bootstrap.go
```diff
@@ -0,0 +1,255 @@
...
+	}()
+}
+
+// setupTracer sets up a root trace for the CLI and initializes the tracer
+func setupTracer(ctx context.Context, name string) context.Context {
```
george-e-shaw-iv, line 98: maybe modify this function to return a closer function that calls `trace.CloseTracer(ctx)`

george-e-shaw-iv: nvm i saw what you did. kinda sketchy but i understand where it comes from with how generic this code has to be

pkg/cli/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package cli contains various cli utilities that are useful for building
+// cli applications with gobox.
+package cli
```
george-e-shaw-iv, line 3:
```suggestion
// Copyright 2021 Outreach Corporation. All Rights Reserved.

// Description: File that houses the package documentation for the cli package.

// Package cli contains various cli utilities that are useful for building
// cli applications with gobox.
package cli
```

pkg/cli/bootstrap.go
jaredallard: yes

george-e-shaw-iv (approved):

pkg/cli/bootstrap.go
```diff
@@ -0,0 +1,262 @@
...
+	if err := trace.InitTracer(ctx, name); err != nil {
+		fmt.Println(err)
+		return ctx
+	}
+	return trace.StartTrace(ctx, name)
+}
+
+// setupPanicHandler sets up a panic handler for CLIs
+func setupPanicHandler(exitCode *int) {
```
vk-outreach, line 105: lol.  I think we can make this a bit more readable by creating a temporary struct to hold all the state and moving all these top-level functions as methods on that struct..

i.e:

```golang

type cli struct {
    exitCode int
    cleanup func()
}

func (c *cli) setupPanicHandler() {
       if r := recover(); r != nil {
          c.exitCode = 2
       }
}
```

PR #81, opened 2021-11-09
feat(region): add API for determining lowest HEAD latency region

jaredallard:

This PR adds a new `pkg/region` library for determining what the best region to use is from the perspective of a client. Provided are the below APIs:

 * `<Cloud>.Regions` - returns all known regions that this cloud has
 * `<Cloud>.Regions.Nearest` - returns the lowest latency region provided by this cloud, this is a readthrough cache that will cache the results of each region for 8 hours
 * `<Cloud>.Regions.Filter` - returns only the provided regions to `Filter`, `Nearest` can be chained onto this.

**JIRA ID**: [DTSS-0]

## Notes for your reviewers

The context for this addition is that we will support multiple GKE clusters and GKE image registries in different regions (EU/US) _very_ soon, this enables that transparently without needing to ask the user or infer based on Geo-IP, which isn't always correct. 8 hours is used as a reasonable assumption for when a good time to re-validate would be e.g. cross-country flight :)


pkg/regionpicker/cache.go
```diff
@@ -0,0 +1,119 @@
+package regionpicker
```
george-e-shaw-iv, line 1:
```suggestion
package region
```

> It's helpful if everyone using the package can use the same name to refer to its contents, which implies that the package name should be good: short, concise, evocative. By convention, packages are given lower case, single-word names

https://golang.org/doc/effective_go#package-names

jaredallard: i like `regionpicker` 😢

vk-outreach: I think `region.Pick` reads a lot better than `regionpicker.FindBest`.   I mean, why would we want to pick anything but the best?  Though I'd probably have chosen `region.Guess` if there is any likelihood of it being wrong or failing.

vk-outreach: Also, `region.Best` and `region.Current` work reasonably well too (don't need to use a verb if we think of it as a getter).

pkg/regionpicker/cache.go
```diff
@@ -0,0 +1,119 @@
...
+
+var (
+	// RegionCachePath is the $HOME/<RegionCachePath> location of the regionpicker cache
+	RegionCachePath = filepath.Join(".outreach", ".cache", "box")
+	// RegionCacheFile is the name of the regionpicker cache file
+	RegionCacheFile = "regionpicker.json"
+)
+
+// cacheEntry is information on a cloud's cache entry
+type cacheEntry struct {
```
vk-outreach, line 21: I think caches should use the standard cache interface we defined [here](https://github.com/getoutreach/datastores/blob/5e7ed55abb4143256326cc95c31401005b37d944/pkg/cache/cache.go#L60).   Maybe we should add a cache.File that backs this implementation onto a single file.

If needed, we can move that interface and implementation into gobox/pkg/cache (and eventually move nearly everything in datastores/pkg/cache over to gobox/pkg/cache).

pkg/regionpicker/cache.go
```diff
@@ -0,0 +1,119 @@
...
+	f, err := os.Create(filepath.Join(cacheDir, RegionCacheFile))
+	if err != nil {
+		return errors.Wrap(err, "failed to create cache file")
+	}
+
+	return json.NewEncoder(f).Encode(c)
+}
+
+// FindCachedBest returns the best region but uses a cache first.
+func FindCachedBest(cloud CloudName, allowedRegions []RegionName, logger logrus.FieldLogger) (RegionName, error) {
```
vk-outreach, line 75: This is an odd API -- why do we need AllowedRegions here? It makes it questionable to write this to a cache as the cache can only be used if the allowedRegions matched the value used originally.

Also, what you are trying to do is a ReadThrough cache for which there is a nice pattern [here](https://github.com/getoutreach/datastores/blob/main/pkg/cache/readthrough.go#L16) (you'd have to implement a cache interface that goes directly against the network and then you can automatically combine that with a cache.File via cache.ReadThrough).

vk-outreach: This is entirely optional but if you went down the route I suggested with cache.Cache interface implementations, there is a very clean way to log cache hits and misses - by wrapping the network call and the cache storage separately (reads from network imply cache miss, writes to storage imply successful fetch etc).

pkg/regionpicker/regionpicker.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// clouds is the global map of all supported clouds
+var clouds = map[CloudName]Cloud{
+	CloudGCP: &GCP{},
+}
+
+// getPing hits an endpoint and returns the ping time
+func getPing(url string) (time.Duration, error) {
```
vk-outreach, line 17: This should really be a method on `region` IMO -- it would also read better in `best` -- `r.Duration(ctx)` ...

pkg/regionpicker/gcp.go
```diff
@@ -0,0 +1,17 @@
+package regionpicker
+
+const (
+	CloudGCP CloudName = "gcp"
+)
+
+type GCP struct{}
+
+// Regions returns a list of all known-gcp regions. This should not be used
+// over native GCP region fetching.
```
deregtd, line 10: this is pretty ugly for now (incomplete).  I'd prolly have users pass in a list for now, if this won't do discovery for them.

pkg/regionpicker/gcp.go
```diff
@@ -0,0 +1,17 @@
...
+// IDEA: One day we could probably get this from GCP. That is authenticated though.
+func (*GCP) Regions() []region {
```
vk-outreach, line 12: I suggest returning a strongly typed exported `Regions` rather than unexported `[]region`.  This would allow you to just move your `best(regions)` helper as a method on `Regions`:

```golang
type Regions []region
func (r Regions) Nearest(ctx context.Context) (RegionName, error) { .... }
```

This would also make your main `FindBest` function read a lot better: `cloud.Regions().Nearest()`

pkg/regionpicker/regionpicker_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"testing"
+
+	"github.com/sirupsen/logrus"
+)
+
+func TestFindBest(t *testing.T) {
+	// should work in general
+	regionName, err := FindBest(CloudGCP, nil, logrus.New())
+	if err != nil {
+		t.Errorf("got an error trying to find best region: Best(): %v", err)
```
vk-outreach, line 13: why not use gotest.tools?  This would read a lot better with it:

`assert.NilError(t, err)`

vk-outreach (commented):
Switching to datastores/pkg/cache may be out of scope here but I suggest considering the interface/naming refactors ..

pkg/regionpicker/cache.go
vk-outreach: Thinking about it a bit more, I think a much better option is to cache all the `ping` durations for the regions rather than `last best time`.  With that approach, you are pushing the cache way down (i.e. your call to `ping` would use the cache and return it if it is there, if it isn't there it would do the actual ping).

The benefit of that approach is that AllowedRegions will not interact with the cache anymore.

pkg/regionpicker/regionpicker_test.go
jaredallard: No particular reason, just hadn't used it for a bit. Can add that

pkg/regionpicker/regionpicker.go
jaredallard: Yeah, fair. Will fix

pkg/regionpicker/cache.go
jaredallard: So the context for "AllowedRegions" is that we need to be able to filter what regions we want to use. For example, we don't want to hit every AZ ever if we're just looking for GCP's multi-region setup `us`, `europe`, and `asia`. 

Feedback on the cache interaction with that is fair, but I'm still somewhat against testing against all regions if we don't need to do so. Will think on this a little bit

pkg/regionpicker/gcp.go
jaredallard: I didn't really want to export `region` because I wanted to protect against usage of this outside of tooling, that being said.... I'm not sure if that's a fair goal to have. So, I'm not super against exporting this then.

vk-outreach: You don't need to expose region -- just Regions and RegionName :)

pkg/regionpicker/cache.go
vk-outreach: I think the last comment I had takes care of avoiding pinging regions that were not request, so let me elaborate a bit:

1. Implement pinger:  `Ping(endpoint) (duration time.Duration, time.Expiration)`
2. Implement a cached pinger which reads from the cache if there is an entry for `endpoint` (and it hasn't expired), falling back to the raw pinger above
3. Implement the picker to use this cached pinger.  So, it always does the `best` calculation but this should be fast if the ping durations are in the cache.

jaredallard: Ahhhh, I see that makes sense!

pkg/region/cache.go
```diff
@@ -0,0 +1,137 @@
...
+// ensure that this is go-routine safe.
+var cache = &cacheStore{}
+
+// cacheEntry is information on a cloud's cache entry
+type cacheEntry struct {
+	// Duration is how long it took to talk to this region at LastUpdatedAt
+	Duration time.Duration `json:"duration"`
+
+	// LastUpdatedAt is the last time this file was updated. UTC.
+	LastUpdatedAt time.Time `json:"last_updated_at"`
```
vk-outreach, line 34: I suggest converting this to ExpiresAt instead.  It is a bit subtle but the main benefit is that you can implement negative caching more naturally eventually (so if a ping failed for a region, you can set duration to be time.10*Hour but set ExpiresAt to be 5 seconds to force it to be cached for far less time).

It doesn't apply in this specific case as much I think but it is worth sticking to that pattern for things that expire.

pkg/regionpicker/cache.go
jaredallard: So, this talks to all the cache stuff. I did this a little differently to allow a global cache in the package (to make it goroutine safe) but also to allow easy of strict typing. If I ever do the work to move datastores/pkg/cache out of that repository then I might re-evaluate, but trying to support that interface without doing that work seemed a little over-optimizing for now and would make this more complex.

PTAL and let me know how you feel about it!

vk-outreach: Haven't reviewed the implementation in detail yet but the interfaces look fine to me.

pkg/region/cache.go
```diff
@@ -0,0 +1,146 @@
...
+
+	mu sync.Mutex
+}
+
+// Get returns the duration of a cloud/region pairing, if it exists.
+// Otherwise ok is returned as false and time.Duration is it's zero-value.
+func (c *cacheStore) Get(cloud CloudName, r Name) (time.Duration, bool) {
+	if c.Clouds == nil {
+		c.load()
+	}
```
george-e-shaw-iv, line 50: you may just want to use a sync.Once for this and the other places you did this

george-e-shaw-iv (approved):
looks good

pkg/region/cache.go
vk-outreach: Yeah, this is not concurrency safe.  You can't even inspect `c.Clouds` without a lock.   I suggest simplifying the whole thing by following these rules:

1. Every exported function just does `c.mu.Lock(); defer c.mu.Unlock(); c.init()` at the top.
2. All unexported functions know the lock is held and can do whatever
3. define `init()` to initialize c.Clouds if it is not already defined.

george-e-shaw-iv: You could also use a `sync.RWMutex` or a `sync.Map` to try and squeeze every little bit of perf out of it if you really cared

jaredallard: good catch, I meant to guard against this

pkg/region/cache.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// ensureKey ensures that a key can be properly accessed in the underlying cache
+func (c *cacheStore) ensureKey(cloud CloudName, r Name) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if _, ok := c.Clouds[cloud]; !ok {
+		c.Clouds[cloud] = make(map[Name]cacheEntry)
+	}
```
vk-outreach, line 71: This isn't needed?  I suggest doing this in Set instead.

```suggestion
```

vk-outreach: More clearly: the first expiration check is only needed for Get (which doesn't need this if block) and this if block is only needed for the Set call (which doesn't need the expiration check).  My suggestion was more along the lines of not combining the two things..

jaredallard: yeah, just realized that, working on it already

pkg/region/region.go
```diff
@@ -0,0 +1,99 @@
...
+
+	startTime := time.Now().UTC()
+	resp, err := http.Head(r.Endpoint) //nolint:gosec // Why: not really variable
+	if err != nil {
+		return 0, err
+	}
+	resp.Body.Close()
+	endTime := time.Now().UTC()
+
+	dur = endTime.Sub(startTime)
```
vk-outreach, line 47: I don't think we need the UTC here (or anywhere for that matter but atleast using UTC makes sense if serialization is involved which isn't the case here).  Also, you can just do `time.Since` : 

```suggestion
  dur := time.Since(startTime)
```

pkg/region/region.go
```diff
@@ -0,0 +1,99 @@
...
+// to do a HEAD request.
+func (r *region) Duration(ctx context.Context) (time.Duration, error) {
+	dur, ok := cache.Get(r.Cloud, r.Name)
+	if ok {
+		return dur, nil
+	}
+
+	startTime := time.Now().UTC()
+	resp, err := http.Head(r.Endpoint) //nolint:gosec // Why: not really variable
+	if err != nil {
```
vk-outreach, line 41: I'm assuming we don't care about HTTP status here?  If so, would be worth adding a comment for it as the most idiomatic thing with HTTP calls is to do `if err != nil || resp.Status > 299` type checks.

jaredallard: We don't care about HTTP status, correct. I'll add that

pkg/region/cache.go
```diff
@@ -0,0 +1,178 @@
...
+
+	_ = json.NewDecoder(f).Decode(&c) //nolint:errcheck // Why: function signature/acceptable
+}
+
+// save saves the cache to disk
+func (c *cacheStore) save() error {
+	c.fileMu.Lock()
+	defer c.fileMu.Unlock()
+
+	cacheFilePath, err := c.getCacheFilePath()
```
george-e-shaw-iv, line 163: why not use the mutex in `getCacheFilePath`? i was always taught to only hold a lock for as long as you need it, and no longer. you're holding this lock for the duration of the function when you really only need it for line 163. you're not doing any more reading after that your locks on 174-175 are correct though

jaredallard: I'm more used to putting locks/unlocks at the top of a body, but you're right it's not needed in there so I'm going to just move it below

vk-outreach: I would favor easy to maintain, correct code over performant code -- atleast until there is a demonstrable reason for perf.  In this case, I'd just use a single mutex and lock at the top of all exported functions -- this is quite easy to see that there is no possibility of deadlocks.  I'd also stick to a single lock per struct for that reason -- very easy to see that the code works correctly.   With multiple locks, code reviews can easily miss correctness problems and concurrency tests are very hard to get right.

jaredallard: Yeah, that's why I tend to do that method, which ensures I don't accidentally mess up locks. That's why I made a `set` and `get` in the follow-up commits earlier. In this case, because I'm calling `defer` right after I'm less concerned about accidentally deadlocking, so I'll keep @george-e-shaw-iv's method for now. Still agree on usually keeping at the top, though.

vk-outreach: For example, we now have two mutexes and a sync.Once in one struct -- I think it is very hard for me to prove that this code is correct.   Even worse, if someoone makes changes, it is easy to miss the interactions between these as the diff will only have the snippet that changed.

I can prove the code is correct from concurrency p.o.v if there was exactly one mutex, all exported functions had the lock/defer-unlock block at the top, unexported functions were only called by exported functions and they had no other locks.

vk-outreach: If you must user multiple locks, I suggest creating embedded structs that scope each lock down:

```golang
type MainStruct struct {
      scope1 struct {
            mu sync.Mutex
            fields protected by mu
      }
      scope2 struct {
            mu sync.Mutex
            fields protected by mu
      }
}
```

This can be validated more easily by confirming that all access to fields within a scope are only acquired with mutexes on that scope.

jaredallard: Updated to use one mutex again and be more clear. I still prefer the `c.once.Do(c.load)` for loading though, that's reading better to me because it's in the "friendly" `Get` and `Set` interfaces only. That work for you?

jaredallard: (reasoning behind going back to one was because they are really tied together for the most part anyways)

jaredallard:
Anything else @vk-outreach? 🚀

pkg/region/cache.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// get returns a cache entry for a given cloud/region pairing
+func (c *cacheStore) get(cloud CloudName, r Name) (*cacheEntry, bool) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	if _, ok := c.Clouds[cloud]; ok {
+		return nil, false
+	}
```
vk-outreach, line 64: nit:

you shouldn't need this either -- the check on L66 should catch it.  [playground example]
(https://play.golang.org/p/IwaZnRNrvYX)

vk-outreach (approved):
LGTM

PR #82, opened 2021-11-11
feat(box): add new fields for multi-cluster/docker image registry support

jaredallard:

This PR adds support to `box` for expressing development docker image registries as well as for listing available clusters for loft. This will be used by the devenv for both `tilt` integration as well as with `loft` for picking and choosing which cluster to use for backing a cloud devenv. This is contingent on me getting direct cluster endpoints to also work, but that's digressing.

**JIRA ID**: [XX-XX]

### DIFF
--- a/pkg/box/box.go
+++ b/pkg/box/box.go
@@ -3,7 +3,9 @@
 // that outreach provides, aka "startup in a box"
 package box
 
-import "time"
+import (
+	"time"
+)
 
 // SnapshotLockChannel is used to determine the quality of
 // a given snapshot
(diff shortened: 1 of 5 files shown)


pkg/box/box.go
```diff
@@ -17,15 +19,15 @@ const (
 	SnapshotLockChannelRC SnapshotLockChannel = "rc"
 
 	// Version is the current version of the box spec.
-	Version float32 = 1
+	Version float32 = 2
```
jaredallard, line 22: Need to bump to force refresh

pkg/box/box.go
```diff
@@ -17,15 +19,15 @@ const (
...
 )
 
 type DeveloperEnvironmentConfig struct {
 	// SnapshotConfig is the snapshot configuration for the devenv
-	SnapshotConfig *SnapshotConfig `yaml:"snapshots"`
+	SnapshotConfig SnapshotConfig `yaml:"snapshots"`
```
jaredallard, line 27: removing pointers because they aren't actually needed

pkg/region/cache.go
```diff
@@ -119,6 +119,10 @@ func (c *cacheStore) getCacheFilePath() (string, error) {
 		storageDir = filepath.Join(homeDir, storageDir)
 	}
 
+	// ensure that the directory exists
+	//nolint:errcheck // Why: best effort
+	os.MkdirAll(storageDir, 0755)
```
jaredallard, line 124: fixes issue with cache not workin

george-e-shaw-iv (approved):

PR #83, opened 2021-11-11
feat: DTSS-976 Add trace.StartSpanAsync

vk-outreach:

Adds a trace.StartSpanAsync to help with async tracing capabilities (when the span does not finish before the parent).
Also adds trace.FromHeadersAsync as a convenience function for use with clerk and workflows.

**JIRA ID**: [DTSS-976]

## Notes for your reviewers

An example honeycomb link is there in the ticket.

[DTSS-976]: https://outreach-io.atlassian.net/browse/DTSS-976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bhogatemadhura (approved):

pkg/trace/trace.go
```diff
@@ -148,6 +148,11 @@ func FromHeaders(ctx context.Context, hdrs map[string][]string, name string) con
 	return defaultTracer.fromHeaders(ctx, hdrs, name)
 }
 
+// FromHeadersAsync fetches trace info from a headers map and kicks off an async trace.
+func FromHeadersAsync(ctx context.Context, hdrs map[string][]string, name string) context.Context {
```
coding-paras, line 152: _Note: I am trying to understand the overall trace as well for the first time._

**Assumption**: Every Golang service sets `t.Honeycomb.Enabled == true` via bootstrapped config. [Here](https://github.com/getoutreach/bootstrap/blob/2cc37917a4666818eb1bec30053e769c3b1afaee/pkg/codegen/templates/deployments/app/app.jsonnet.tpl#L118)

AFAIU, the clerk workflow would look like follows:

**_Use-case: Span across 3 hops_**
1: Source producer
2: Re-publisher - Consume from 1, Produce for 3
3: Final consumer

**_My understanding:_** The name of the trace needs to be unique to help honeycomb to stitch things.

**_Flow_**
Code is mix of pseudo and real

**_[1] Source producer_**
```
ctx := // service's context

// Get trace headers to start async span
headers := trace.ToHeaders(ctx)

// Generate a unique name for the trace at source
traceName := "blah-blah-blah-event-42"

// Start child trace
sourceProducerTrace := trace.FromHeadersAsync(ctx, headers, traceName)
trace.AddInfo(sourceProducerTrace, log.F{"trace": traceName})

// Generate and produce clerk evt
evtCtx := // clerkContextObject
evtCtx.TraceName = traceName // We will add this field
evt := createNewClerkEvt(evtCtx, evt)
clerk.Produce(ctx, evt)

// DO NOT end it since some other process/pod will end this
```

**_[2] Re-publisher - Consume from 1, Produce for 3_**
```
// Clerk Consumer handler code

prevEvt := // evt from clerk
prevTraceName := evt.EvtCtx.TraceName

ctx := // service's context

// Get trace headers to start async span
headers := trace.ToHeaders(ctx)

// Start child trace
republisherTrace := trace.FromHeadersAsync(ctx, headers, prevTraceName)
trace.AddInfo(republisherTrace, log.F{"trace": prevTraceName})

// Generate and produce clerk evt again
evtCtx := // clerkContextObject
evtCtx.TraceName = prevTraceName
evt := createNewClerkEvt(evtCtx, evt)
clerk.Produce(ctx, evt)

// DO NOT end it since some other process/pod will end this
```

**_[3] Final consumer_**
```
// Clerk Consumer handler code

prevEvt := // evt from clerk
prevTraceName := evt.EvtCtx.TraceName

ctx := // service's context

// Get trace headers to start async span
headers := trace.ToHeaders(ctx)

// Start child trace
finalConsumerTrace := trace.FromHeadersAsync(ctx, headers, prevTraceName)
trace.AddInfo(finalConsumerTrace, log.F{"trace": prevTraceName})


// End trace
trace.End(finalConsumerTrace)
```

If the above assumption is true, does every grpc/http call get a unique name that we can extract via some static function? Then we can just attach this span to the grpc/http call itself.

coding-paras: Reading this thing again - I feel like I am confused between `traceIds` and `traceNames`.

vk-outreach: A trace ID is the actual unique identifier -- the name is just a human readable category for the trace.  For gRPC calls, we use the gRPC method name as the name of the trace.

coding-paras: So in my above pseudo-code, do I have to pass `traceId` somewhere?

vk-outreach: The trace ID is the overall correlation ID.   Everything processing one message would have this same trace ID.  In addition, each node (span in trace terminology) would have its own ID and an id of the parent (so you can build the call tree).  All of these are serialized into a set of http headers by the call trace.ToHeaders.

Here is the clerk call flow: https://gist.github.com/vk-outreach/1533239ddb9a9d504b26c852010c9d74

The example above has two changes: it omits names altogether from clerk and instead uses TraceHeaders. Note also the publisher side has nothing special going on about async spans -- it is only on the consumer side.

Re: the trace headers -- if we plan to use our data lake to do analysis of our call trees, it might make sense to change the clerk trace context to explicitly list the trace related fields as `Trace ID, Span ID, Span Type, Parent SpanID` instead of the opaque TraceHeaders.

coding-paras: I got the answer. We need to pass `traceheaders`. Thank you.

coding-paras (approved):

PR #87, opened 2021-11-16
fix(resources): change time values to pointers

nissim-natanov-outreach:
Apparently, k8s does not handle well its own dateTime zero time values. Converter changes zero Time{} to "null" string - and from some unknown to me reason, K8S itself fails to accept 'null' as a valid dateTime value. This PR attempts to use ptr instead with omitempty as an attempt to fix this issue.

**JIRA ID**: [QSS-00]


nissim-natanov-outreach:
apparently, this fix is no longer needed, close it. the root cause of errors were incorrect definition of the dateTime in CRD:
wrong: { type: 'dateTime' }
right: {type: 'string', format='date-time' nullable: true}

PR #90, opened 2021-11-18
feat(resources): move Reconciler impl to gobox

nissim-natanov-outreach:
This PR moves reconciler logic off bootstrap into gobox, and implements Requeue logic there. It also adds unit tests.

**JIRA ID**: [QSS-799]

## Notes for your reviewers

[QSS-799]: https://outreach-io.atlassian.net/browse/QSS-799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/k8s/controllers/reconciler.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+// Reconcile is invoked when controller receives resource spec.
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := r.log.WithField("resourceName", req.NamespacedName).WithField("kind", r.Kind())
+	rr := r.doReconcile(ctx, log, req)
+
+	// defer as func to avoid 'capturing' rr - we want to provide latest values to the endReconcile
+	defer func() {
+		r.endReconcile(ctx, log, *rr)
```
vk-outreach, line 142: Strictly speaking, we don't actually need this callback setup at all.  Any service which uses this package can actually do their `post reconcile` activity by just embeddding:

```golang

type MyReconciler struct {
    // embed controller's Reconciler
    controllers.Reconciler
}
func (my *MyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
    result, err := my.Reconciler.Reconciler(ctx, req)
    // do endReconcile work on top of result, err here.
}
```

I suggest just dropping EndReconcile out of the Handler interface altogether and this defer also.  If we really want to share the code of how to log the result, I suggest adding a helper to Reconciler to lot the result:

```golang
func (r *Reconciler) LogResult(ctx context.Context, req ctr.Result, res ctrl.Result, err error) {
}
```

nissim-natanov-outreach: endReconcile carries skip logic - which is not available as a return value. the built-in version delegates to Handler.EndReconcile and also log full state, and it needs skip

vk-outreach: Maybe add skip to `Reconciler.Reconciler` ?  Another option is to return a custom error for skip.

nissim-natanov-outreach: also, Handler's Reconcile is not always called - while EndReconcile is always invoked.

vk-outreach: Whether handler's Reconcile gets called or not doesn't really matter as I'm suggesting the bootstrapped service actually wraps `Reconciler.Reconcile` -- ie. wherever `controllers.NewReconciler()` is called, it would be replaced with `&MyReonciler{controllers.NewReconciler()}`.

nissim-natanov-outreach: I could modify the Reconcile signature (make it sort of a helper utility) - but then controller must implement the interface. I wanted to keep bootstrapped code with 'code' in it to minimum, relying on Reconciler to be the primary receiver of the callback, while Handler gets all for free. EndReconcile is a way for the handler to handle success and error cases, and errors may come from the very beginning too (e.g. failed to even read CR from the k8s).

This way we can keep bootstrapped code to bare min, only asking for the actual core implementation.

evadnoob: Isn't skip logic trivial to implement by returning `return ctrl.Result{}, nil` from Reconcile, I'm not sure. I concur with @vk-outreach, returning out of `Reconcile` is more consistent with the api and generated code from kubebuilder(which I fear we have strayed far from at this point).

pkg/k8s/controllers/reconciler.go
```diff
@@ -0,0 +1,276 @@
...
+// Reconcile is invoked when resource spec is created or updated.
+func (r *Reconciler) doReconcile(
+	ctx context.Context,
+	log logrus.FieldLogger,
+	req ctrl.Request,
+) *ReconcileResult {
+	rr := ReconcileResult{}
+	in := r.handler.CreateResource()
+	if getErr := r.Client().Get(ctx, req.NamespacedName, in); getErr != nil {
+		log.WithError(getErr).Errorf("unable to get %s CR", r.Kind())
```
evadnoob, line 179: The error here can also happen if the custom resource is deleted, I believe, but I'd need to test that case to be sure.

pkg/k8s/controllers/reconciler.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	rr = r.handler.Reconcile(ctx, log, in)
+	if rr.Skipped {
+		// do not take any action here if impl asked to skip status+hash updates (maybe CR is meant for a diff bento)
+		// endReconcile is still called - logging done inside
+		return &rr
+	}
+
+	updateErr := r.updateStatus(ctx, log, in, &rr)
```
evadnoob, line 210: What happens if the custom resource doesn't have a status sub-resource?

pkg/k8s/controllers/reconciler.go
```diff
@@ -0,0 +1,276 @@
+// Copyright 2021 Outreach Corporation. All Rights Reserved.
+
+// Description: This file defines a kubernetes controller for v1/PostgresqlDevenvDatabase.
```
deregtd, line 3: Shouldn't this ... not be in gobox?

pkg/k8s/controllers/reconciler.go
nissim-natanov-outreach: " return ctrl.Result{}, nil" is not skip - it is success, and we update LastReconcileSuccessHash on success. Skip means - do not update anything - just ignore the CR altogether because it belongs to a different bento or because it has been already applied.

pkg/k8s/controllers/reconciler.go
nissim-natanov-outreach: hmm, then why would controller infra call Reconcile if CR is deleted?

pkg/k8s/controllers/reconciler.go
nissim-natanov-outreach: we want all resources using this framework to use Status - bootstrap forces it too

pkg/k8s/controllers/reconciler.go
nissim-natanov-outreach: doh, nice catch -  I will remove copy-pasted code soon

pkg/k8s/controllers/reconciler.go
evadnoob: It's something to test, I have a vague recollection of this being the case.

> A Kubernetes Controller is a routine running in a Kubernetes cluster that watches for create / update / delete events on Resources

Controllers get `Reconcile` when resources are deleted, so they can cleanup or respond to delete actions.

pkg/k8s/controllers/reconciler.go
```diff
@@ -0,0 +1,276 @@
...
+
+	"github.com/getoutreach/gobox/pkg/k8s/resources"
+	"github.com/sirupsen/logrus"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// requeueIntervals are the intervals with which we will requeue failed CRs
+// last value (also available as maxRequeueInterval) is used after we tried all
+var requeueIntervals = []time.Duration{
```
evadnoob, line 18: These seem restrictive, what if a controller wants to retry in 100ms? is it possible to override these?

pkg/k8s/controllers/reconciler.go
nissim-natanov-outreach: updated now, thanks

pkg/k8s/controllers/reconciler.go
nissim-natanov-outreach: oh, ok - prob needs to handle this particular error somehow by detecting CRD delete case.

nissim-natanov-outreach: Let's chat about this offline since I am not sure what is the priority of it. Shall we just ignore the CR for now if deleted since we do not support it anywhere yet anyway?

pkg/k8s/controllers/reconciler.go
nissim-natanov-outreach: yes - controller can ask for a different requeue - and Reconciler will honor it. I added code for this - see endReconcile.

pkg/k8s/resources/resource.go
```diff
@@ -0,0 +1,16 @@
...
+// Hash is used to determine if spec has changed since the last status update or not.
+type ResourceSpec interface {
+	Hash() (string, error)
+}
+
+// Resource represents the custom resource object.
+type Resource interface {
+	client.Object
+	GetSpec() ResourceSpec
+	GetStatus() *ResourceStatus
```
vk-outreach, line 15: It is more idiomatic to name getters without the Get in Golang. Reference: [Effective Go](https://golang.org/doc/effective_go#Getters)

```suggestion
	Spec() ResourceSpec
	Status() *ResourceStatus
```

pkg/k8s/resources/resource.go
```diff
@@ -0,0 +1,16 @@
+package resources
+
+import "sigs.k8s.io/controller-runtime/pkg/client"
+
+// ResourceSpec allows generic Reconciler's implementation to get Hash value of the CR's Spec.
+// Hash is used to determine if spec has changed since the last status update or not.
+type ResourceSpec interface {
```
vk-outreach, line 7: I suggest just naming this Spec (& similarly for Status), this is more idiomatic in Golang and reads better when you type `resources.Spec` as opposed to `resources.ResourceSpec`.

pkg/k8s/controllers/reconciler.go
```diff
@@ -0,0 +1,276 @@
...
+		For(r.handler.CreateResource()).
+		Complete(r)
+	if err != nil {
+		r.log.WithError(err).Errorf("failed to setup Reconciler as a controller for %s resource with k8s manager", r.Kind())
+	}
+	return err
+}
+
+// getRequeueDuration returns the requeue interval to retry, based on number of times this CR failed so far.
+func getRequeueDuration(failCount int) time.Duration {
```
vk-outreach, line 263: golang naming convention: s/get//g

```suggestion
func requeueDuration(failCount int) time.Duration {
```

pkg/k8s/resources/resource.go
nissim-natanov-outreach: Sorry, it collides with existing Spec and Status fields. Unfortunately, Go does not allow method and field names to be the same

pkg/k8s/controllers/reconciler.go
nissim-natanov-outreach: if controller does not need status, it prob does not need the framework here as well. I would prefer status to be there, even though you do not know if it is needed or not. it helps troubleshooting a lot!

pkg/k8s/controllers/reconciler.go
nissim-natanov-outreach: ack, will push shortly

pkg/k8s/resources/resource.go
nissim-natanov-outreach: The package is 'controllers' so it is not obvious which spec is that. Also, I already have ResourceStatus - so it kinda consistent now.

pkg/k8s/controllers/reconciler.go
nissim-natanov-outreach: done

george-e-shaw-iv (approved):

PR #92, opened 2021-11-19
fix: upgrade honeycomb to get a bump in klaus compress lib

ChrisPRobinson:
Upgrade honeycomb library to get an improved klaus/compress library with zstd optimizations

**JIRA ID**: [XX-XX]


ChrisPRobinson:
Dependabot already made this PR

PR #93, opened 2021-11-19
feat(reconciler): handle CR Not Found case

nissim-natanov-outreach:
Delegate Not Found errors to the handler via the new Deleted callback, and add unit test

**JIRA ID**: [QSS-822]

## Notes for your reviewers

[QSS-822]: https://outreach-io.atlassian.net/browse/QSS-822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/k8s/controllers/reconciler.go
+++ b/pkg/k8s/controllers/reconciler.go
@@ -48,15 +48,15 @@ type Handler interface {
 		in resources.Resource) ReconcileResult
 
 	// EndReconcile is called when reconciliation finishes. It is always called, even if reconcile fails before calling
-	// the Handler's Reconcile method.
+	// the Handler's Reconcile method and also if resource is Not Found (with NotFound flag = true).
 	// This method is for logging and metrics, ReconcileResult is intentionally passed by value so there is no point modifying it.
 	EndReconcile(ctx context.Context, log logrus.FieldLogger, rr ReconcileResult)
 
-	// Deleted callback is called when resource is detected as Not Found. Be careful handling deleted database objects as they
+	// NotFound callback is called when resource is detected as Not Found. Be careful handling deleted database objects as they
 	// can lead to accidental and total data loss!
-	// Note: Deleted callback does not have to set Deleted flag on the result (although no harm doing so because
-	// the infra will set it on ReconcileResult right after invoking the Deleted callback anyway).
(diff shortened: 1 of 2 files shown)


pkg/k8s/controllers/reconciler.go
```diff
@@ -124,6 +137,11 @@ func (r *Reconciler) endReconcile(
 		return
 	}
 
+	if rr.Deleted {
```
evadnoob, line 140: I think we should differentiate between "Not Found" or "other error", Not found could mean deleted.

nissim-natanov-outreach: see below - r.Deleted is set ONLY on Not Found

nissim-natanov-outreach: I also have a test for it - one for Not Found and another for 'any other err'

nissim-natanov-outreach: I can rename to NotFound ...

george-e-shaw-iv (approved):
this looks fine just make sure you're being careful with your interface member functions and the interface in general. they're very hard to change once people have started using them unless you design them to be future-proof

if you pay attention in the stdlib there are barely any interfaces bigger than a few member functions. they do that intentionally and then make composite interfaces out of them

nissim-natanov-outreach:
> this looks fine just make sure you're being careful with your interface member functions and the interface in general. they're very hard to change once people have started using them unless you design them to be future-proof
> 
> if you pay attention in the stdlib there are barely any interfaces bigger than a few member functions. they do that intentionally and then make composite interfaces out of them

agree - those are likely the most common cases. For other corner cases we would split logic into New interface and do type checks. Since DPO is the only operator using this for now, adding new core logic is fine.

pkg/k8s/controllers/reconciler.go
nissim-natanov-outreach: renamed Deleted to NotFound - although technically there is no case in which k8s would call Reconcile on never existing CRs :)

evadnoob (approved):
adds NotFound handing, this is in sync with the api to me, LGTM

PR #96, opened 2021-11-23
fix(reconciler): fix endless retry on error

nissim-natanov-outreach:
During devenv testing I found that reconciler will endlessly retry with no delays if error is reported back, e.g. requeue intervals were not honored. This is due to status update unconditionally scheduling another reconcile event, and controller's requeue return value was not taking effect at all.

To fix the above, I introduced NextReconcileTime to enforce retry schedule. This way:
* if NextReconcileTime is set, it will be enforced (ShouldReconcile changed to support this)
* if NextReconcileTime is not set, the error is deemed to be permanent and reconciliation process will stop. This is rare and not used yet by DPO - but we might one do do so. To report permanent error, reconciler must set both Skipped and ReconcileErr (e..g making it fail and skip further retries). Comments updated to make usage clear.

Unit tests updated too.

**JIRA ID**: [QSS-00]


pkg/k8s/controllers/reconciler.go
```diff
@@ -14,28 +13,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// requeueIntervals are the intervals with which we will requeue failed CRs
```
nissim-natanov-outreach, line 17: requeue interval/functions moved to a separate file

pkg/k8s/controllers/requeue.go
```diff
@@ -0,0 +1,41 @@
+package controllers
```
george-e-shaw-iv, line 1: could you not use an exponential backoff algorithm here - see the retry package that is either in services or this repo, can't remember

george-e-shaw-iv (approved):
it looks good but consider using exponential backoff as opposed to static retry times

pkg/k8s/controllers/requeue.go
nissim-natanov-outreach: added, thanks!

george-e-shaw-iv (approved):

pkg/k8s/controllers/errors.go
```diff
@@ -0,0 +1,35 @@
...
+
+// Permanent wraps an existing error to make it permanent (e.g. no more retries allowed).
+func Permanent(err error) error {
+	return PermanentError{wrappedError: wrappedError{Original: err}}
+}
+
+// PropagateError wraps an existing error to mark it as propagatable to controller. If Reconciler's handler
+// returns wraps the error with this struct, Reconciler unwraps it and return it to the controller infra.
+// Use Propagate method to create it.
+type PropagateError struct {
```
vk-outreach, line 27: The recommended wrapping pattern is to to implement Unwrap (see [here](https://github.com/getoutreach/gobox/blob/ffba60cecdc843119e94ff39211558005d852333/pkg/orerr/orerr.go#L126)).

This will ensure this plays well with Go's [error treatment](https://go.dev/blog/go1.13-errors).

Also, I do not recommend exporting the concrete error types and instead exporting `IsXYZ` method (much like with [IsRetryable](https://github.com/getoutreach/gobox/blob/ffba60cecdc843119e94ff39211558005d852333/pkg/orerr/orerr.go#L136)).  This will allow changing the underlying implementation later.

In fact, could you just use IsRetryable to mark errors are retryable?  Do you even need the Propagate and Permanent error types?

pkg/k8s/controllers/reconciler.go
```diff
@@ -142,48 +134,22 @@ func (r *Reconciler) endReconcile(
...
-		// if we do not report error to the controller, we need to requeue for this CR
-		if rr.ControllerRes.Requeue || rr.ControllerRes.RequeueAfter > 0 {
-			// Reconcile alredy asked for requeue, honor it
-			return rr.ControllerRes, nil
-		}
-
-		// Reconcile reported an error and did not ask for a requeue.
-		// This is OK, we will requeue with incrementing intervals.
-		rr.ControllerRes.RequeueAfter = requeueDuration(rr.failCount)
+	propagateErr, ok := rr.ReconcileErr.(PropagateError)
```
vk-outreach, line 149: Please do not cast errors to concrete types.  Use [errors.Is](https://pkg.go.dev/errors#Is).  See [this](https://docs.google.com/document/d/1IsO9CfMpkwEJRKRmexvjKtb8XlOVXR_Ja8l99OG4JtM/edit#heading=h.8swsdcpfjiyf).

pkg/k8s/controllers/errors.go
nissim-natanov-outreach: Oh, I am going to close this PR - code moved to services repo. I will apply comments in the new repo later, thanks!!

pkg/k8s/controllers/reconciler.go
nissim-natanov-outreach: good point, I will apply it to the new repro soon.

nissim-natanov-outreach:
We have moved k8s code from gobox to services. I am going to apply comments there soon.

PR #99, opened 2021-12-02
fix: Use not-authenticated http client for downloading updates

plisy:
The URL you can see on release pages on GitHub redirects to an already authenticated URL (has auth as query params).

This endpoint doesn't allow double auth, and the github client's http client automatically adds Bearer auth header.

Using default http client, we avoid this issue.

**JIRA ID**: [XX-XX]

## Notes for your reviewers
You can see the issue being fixed here on  bootstrapbot check.

### DIFF
--- a/pkg/updater/github.go
+++ b/pkg/updater/github.go
@@ -252,12 +252,15 @@ func (g *Github) DownloadRelease(ctx context.Context, r *github.RepositoryReleas
 		return "", func() {}, err
 	}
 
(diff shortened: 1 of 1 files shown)


deregtd (approved):

huazhoul-outreach (approved):

plisy:
> Does this actually cause issues? I'd imagine this wouldn't have ever worked if it did. That being said I'm OK with changing it, just confused what prompted it.

It started yesterday. I guess they added a check now.

PR #106, opened 2021-12-21
feat(log): add logr adapter

jaredallard:

This PR adds an adapter for `gobox/pkg/log` that implements the `go-logr/logr.Sink` interface.

**JIRA ID**: [XX-XX]

### DIFF
--- a/pkg/log/adapters/logr.go
+++ b/pkg/log/adapters/logr.go
@@ -1,0 +1,107 @@
+// Copyright 2021 Outreach Corporation. All Rights Reserved.
+
+// Description: This file integrates the logger with go-logr/logr
+package adapters
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/getoutreach/gobox/pkg/events"
+	"github.com/getoutreach/gobox/pkg/log"
+	"github.com/go-logr/logr"
+)
+
+// NewLogr returns a gobox/pkg/log logger that implements the
+// logr.Logger interface.
+//
+// ! This should ONLY be used if the consumer doesn't support calling
+// ! gobox/pkg/log directly.
+func NewLogrLogger(ctx context.Context) logr.Logger {
+	logger := &logrLogger{ctx, []log.Marshaler{}}
+	return logr.New(logger)
+}
+
+// logrLogger implements logr.Logger
+type logrLogger struct {
+	ctx context.Context
+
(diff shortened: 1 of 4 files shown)


george-e-shaw-iv (approved):

pkg/log/adapters/logr.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// Errors wraps log.Error()
+func (l *logrLogger) Error(err error, msg string, keysAndValues ...interface{}) {
+	log.Error(l.ctx, msg, append(append(l.existingMarshalers, events.Err(err)), listToGoboxF(keysAndValues...))...)
+}
+
+// Info wraps log.Info()
+func (l *logrLogger) Info(level int, msg string, keysAndValues ...interface{}) {
+	log.Info(l.ctx, msg, append(l.existingMarshalers, listToGoboxF(keysAndValues...))...)
```
vk-outreach, line 90: there is a subtle bug here with `append()` -- it mutates the underlying buffer and more importantly it is not thread-safe.   Just don't use append if you can help it :)

In this case, I suggest not doing any appends at all and just using log.Many like so:

`log.Info(l.ctx, msg, log.Many(l.existingMarshaler), log.Many(listtoGoboxF(...)))`

though if you chagne the types of l.existingMarshaler to log.Many, you can just pass it directly

pkg/log/adapters/logr.go
```diff
@@ -0,0 +1,107 @@
...
+func (l *logrLogger) WithName(name string) logr.LogSink {
+	return l
+}
+
+// WithValues returns a copy of the current logger with the provided
+// key/value pairs being added to all sub-sequent calls
+// of error/info/etc.
+func (l *logrLogger) WithValues(keysAndValues ...interface{}) logr.LogSink {
+	newLogger := &logrLogger{
+		existingMarshalers: append(l.existingMarshalers, listToGoboxF(keysAndValues...)),
```
vk-outreach, line 104: The append here is also a problem.  If you change existingMarshalers to be `log.Many`, you can just do:

```golang
    return &logrLogger{existingMarshalers: log.Many{l.existingMarshalers, listToGoboxF(keysAndValues...)}}
```

nissim-natanov-outreach: @jaredallard tagging you just in case ...

PR #113, opened 2022-01-13
fix(cli): dont panic if no fallback reader is set

jaredallard:

Fixes hard to debug issues in CLIs. Still figuring out the true CLI issue.

**JIRA ID**: [XX-XX]


evadnoob (approved):
improves failure case, nil fallbackSecretLookup, LGTM

PR #118, opened 2022-01-19
feat: trace ids in logs [DRAFT]

kadukf-outreach:
It is required to pass trace IDs (trace id, span id, parent id) into all logs. 

**JIRA ID**: no jira, just proposal for a discussion yet

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -129,19 +129,19 @@ func (f F) MarshalLog(addField func(field string, value interface{})) {
 // Debug emits a log at DEBUG level but only if an error or fatal happens
 // within 2min of this event
 func Debug(ctx context.Context, message string, m ...Marshaler) {
-	dbgEntries.Append(format(message, "DEBUG", time.Now(), app.Info(), m))
+	dbgEntries.Append(format(ctx, message, "DEBUG", time.Now(), app.Info(), m))
 }
 
 // Info emits a log at INFO level. This is not filtered and meant for non-debug information.
 func Info(ctx context.Context, message string, m ...Marshaler) {
-	s := format(message, "INFO", time.Now(), app.Info(), m)
+	s := format(ctx, message, "INFO", time.Now(), app.Info(), m)
 
 	Write(s)
 }
 
 // Warn emits a log at WARN level. Warn logs are meant to be investigated if they reach high volumes.
 func Warn(ctx context.Context, message string, m ...Marshaler) {
-	s := format(message, "WARN", time.Now(), app.Info(), m)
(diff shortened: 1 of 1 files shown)


pkg/log/log.go
```diff
@@ -207,3 +210,8 @@ func generateFatalFields(entry F) {
 	entry["error.message"] = "fatal occurred"
 	entry["error.stack"] = string(debug.Stack())
 }
+
+var appenders = make([]func(ctx context.Context) Marshaler, 0)
```
vk-outreach, line 214: I think this is awkward.  Why not just change trace as our default logging package as well (and have it call log with the current trace info added) ?

i.e. we extend trace to include all the default log functions but it also adds the trace info.  

We would need to change all shared library and related code to use trace instead of log but that seems relatively simple.  Or we could just make log.XXX point to trace.XXX.

vk-outreach: Another approach:

1. move all logging functionality into pkg/logger but don't use globals (instead require log.New).
2. move trace to define a DefaultLogger and switch to using that.  And also extend trace to implement the log.XXX equivalents by automatically adding the trace Info.
3. point log.XXX to use trace.XXX instead.

vk-outreach: We already polluted trace with trace.Call business (that includes logging) and it might be time to just admit that tracing and logging are too coupled together for us.

jaredallard:
Please leave a comment in the description of the PR, this is an open-source repository and we should ensure that we're communicating changes on it. Thanks!

kadukf-outreach:
@jaredallard I'm sorry, this was just a draft to start brainstorming with VK, thanks for marking it as draft. I'll do better next time with drafts, too

pkg/log/log.go
kadukf-outreach: Yeah. I had various versions but wanted to create minimal change. I see it following;
Current situation:
image
IMHO too messy, no need to say that in tracing we mix plenty of other stacks (http headers, rountrippers, etc) All this needs its own place.Back to the topic. 

Logging, tracing and metering serve its own purpose. We want to combine some of the aspects (logging+tracing) in some cases so that developers can rely on the fact. that tracing of the calls (btw, same as span just differs by logging/metering aspect) emits log+trace+meter. IMO, metering. can be solved by log-to-metrics so I'd like to skip it for now. 

The topic is too broad here for the PR, I'm preparing a document discussing it but until that, solving tracing ID in the logs with minimal change - what about pushing "trace ID info" to the context and let the logger use the logging context and emit it. **Internal** for now until we solve log context correlation topic.

PR #121, opened 2022-01-21
fix: Refactor pkg/trace and pkg/log a bit by pulling out some helpers into internal/

vk-outreach:
This refactors the trace package which has gotten really unwieldy.
**JIRA ID**: [XX-XX]

There is no functional change.  Just pulling out some functionality into internal packages. 

I pulled out the recursive marshaling functionality into logf (as well as log.F and log.Many) allowing us to share this between trace and log.
I pulled out the core call tracking functionality into `call`.  

I had more changes but the PR was getting larger, so I decided to shave this off by itself first.

### DIFF
--- a/internal/call/call.go
+++ b/internal/call/call.go
@@ -1,0 +1,119 @@
+// The call package helps support tracking latency and other metrics for calls.
+package call
+
+import (
+	"context"
+	"time"
+
+	"github.com/getoutreach/gobox/internal/logf"
+	"github.com/getoutreach/gobox/pkg/app"
+	"github.com/getoutreach/gobox/pkg/events"
+	"github.com/getoutreach/gobox/pkg/metrics"
+)
+
+// Type tracks the call type.
+type Type string
+
+const (
+	// TypeHTTP is a constant that denotes the call type being an HTTP
+	// request.
+	TypeHTTP Type = "http"
+
+	// TypeGRPC is a constant that denotes the call type being a gRPC
+	// request.
+	TypeGRPC Type = "grpc"
+
+	// TypeOutbound is a constant that denotes the call type being an
+	// outbound request.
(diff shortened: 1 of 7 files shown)


internal/call/call.go
```diff
@@ -0,0 +1,119 @@
...
+	Type Type
+	Kind metrics.CallKind
+	Args []logf.Marshaler
+
+	events.Times
+	events.Durations
+
+	ErrInfo *events.ErrorInfo
+}
+
```
jaredallard, line 43: All of these need comments on them because lintroller is going to be ran on this soon 😓

pkg/trace/call.go
```diff
@@ -3,88 +3,16 @@ package trace
...
-func (c *callInfo) ReportGRPCLatency(err error) {
-	metrics.ReportGRPCLatency(app.Info().Name, c.name, c.ServiceSeconds, err, metrics.WithCallKind(c.kind))
-}
-
-// ReportOutboundLatency is a thin wrapper around metrics.ReportOutboundLatency to report latency
-// metrics for outbound calls.
-func (c *callInfo) ReportOutboundLatency(err error) {
-	metrics.ReportOutboundLatency(app.Info().Name, c.name, c.ServiceSeconds, err, metrics.WithCallKind(c.kind))
-}
+// nolint:nochecknoglobals
```
jaredallard, line 14: Needs a `// Why: ....`

internal/call/call.go
vk-outreach: Ah great.  Fixed now.

pkg/trace/call.go
vk-outreach: Fixed.

jaredallard (approved):

PR #124, opened 2022-01-25
feat(trace): add scheduled time setter

alexandrmaximov-outreach:
We'd like to be able to set scheduled time to have a correct waiting and total time in logs.

**JIRA ID**: [XX-XX]

### DIFF
--- a/internal/call/call.go
+++ b/internal/call/call.go
@@ -86,6 +86,15 @@ func (info *Info) AddArgs(ctx context.Context, args ...logf.Marshaler) {
 	info.Args = append(info.Args, args...)
 }
 
+// ApplyOpts applies CallOption functions to the call Info object.
+func (info *Info) ApplyOpts(ctx context.Context, args ...logf.Marshaler) {
+	for _, a := range args {
+		if opt, ok := a.(CallOption); ok {
+			opt(info)
+		}
+	}
+}
+
 // SetStatus updates the ErrInfo field based on the error.
 func (info *Info) SetStatus(ctx context.Context, err error) {
 	info.ErrInfo = events.NewErrorInfo(err)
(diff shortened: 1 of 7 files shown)


pkg/trace/call.go
```diff
@@ -72,6 +73,13 @@ func SetCallTypeOutbound(ctx context.Context) context.Context {
 	return ctx
 }
 
+// SetCallScheduled is meant to set the call scheduled at time on a context that
+// has already been initialized for tracing via StartCall or StartExternalCall.
+func SetCallScheduled(ctx context.Context, scheduled time.Time) context.Context {
```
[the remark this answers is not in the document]
alexandrmaximov-outreach: Option 2 seems to be better. Should we do a new major release with breaking change or add another `StartCallSmth` method to allow both old and new API simultaneously?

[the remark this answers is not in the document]
vk-outreach: I think the way I wrote it up, it won't be a breaking change since `CallOption` implements log.Marshaler.  Not pretty but it is effective.

vk-outreach (approved):
LGTM

internal/call/call.go
```diff
@@ -86,6 +86,15 @@ func (info *Info) AddArgs(ctx context.Context, args ...logf.Marshaler) {
 	info.Args = append(info.Args, args...)
 }
 
+// ApplyOpts applies CallOption functions to the call Info object.
+func (info *Info) ApplyOpts(ctx context.Context, args ...logf.Marshaler) {
+	for _, a := range args {
+		if opt, ok := a.(Option); ok {
```
george-e-shaw-iv, line 92: this is such a better interface than wrapping function calls all centered around a context. the only thing i'm a little lost on is why pass args as all types that implement the `logf.Marshaler` interface when what you actually want is only one concrete type - the `Option` interface. Is that just forwards thinking? if so, I think its worth a comment because it's not immediately obvious

george-e-shaw-iv (approved):

internal/call/call.go
george-e-shaw-iv: yeah i probably should've read VKs comment first. that makes sense. comment is a nit, if you think it's unnecessary thats fine w/me

alexandrmaximov-outreach: Sure, I've added godoc comment

PR #128, opened 2022-02-02
feat(github): expose github auth helpers

jaredallard:

This PR adds a new `github` package that exposes a `GetToken` function that supports all known auth methods for Github token's on user's local machines. This was inspired by the original package in stencil `pkg/extensions/github`

**JIRA ID**: [DTSS-0]


ChrisPRobinson (approved):
LGTM

PR #132, opened 2022-02-04
feat(null): import getoutreach/null

zilwang:
import getoutreach/null which provides support for setting values explicitly to null

**JIRA ID**: [OLM-1503]

## Notes for your reviewers

[OLM-1503]: https://outreach-io.atlassian.net/browse/OLM-1503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jaredallard:
Hey, can we get some context on why this is being added? Aliasing a package into gobox is somewhat a big deal but also in general for our stdlib we want to make sure we're comfortable before adding something. Thanks!

zilwang:
> Hey, can we get some context on why this is being added? Aliasing a package into gobox is somewhat a big deal but also in general for our stdlib we want to make sure we're comfortable before adding something. Thanks!

Hi, this is a part of changing our go `apiv1` client to support `omitempty` and allowing us to leverage partial update functionality in the flagship.
Full details are a bit long, but here is the link to the thread https://outreach-hq.slack.com/archives/CM3TW594L/p1644000142986199

pkg/null/null.go
```diff
@@ -0,0 +1,16 @@
...
+// separate values, use these.
+package null
+
+import "github.com/getoutreach/null"
+
+type Bool = null.Bool
+type Float = null.Float
+type Int = null.Int
+type String = null.String
+type Time = null.Time
```
george-e-shaw-iv, line 16: I don't understand the need for this intermediate layer. What's wrong with importing the fork directly wherever you need to use it?

PR #134, opened 2022-02-09
feat: add experimental tracelog provider

vk-outreach:
Adds an experimental trace logger which can log all info available in the call context.

**JIRA ID**: [DTSS-795]

## Notes for your reviewers
This is a prototype of the proposal [described in this doc](https://docs.google.com/document/d/1PFmV5pfgiKxRS4SDaGZqCFu6Qf4f78nM75bHM72XlyY/edit) but without any invasive refactoring of pkg/trace itself (as that can be done separately).

This change restructures pkg/trace to allow adding an experimental trace logger which can log extra info in trace.StartCall/trace.EndCall.

Note that the experimental tracelogger itself is in x/tracelog and as such we can iterate on it without backwards compatibility issues.

Also note while this change adds trace.AddProvider, there is no major refactoring of trace itself as that is deferred to separate PR.


pkg/trace/providers.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+}
+
+var providers = manyProviders{&defaultCallProvider{}}
+
+// AddProvider is an experimental interface to add a provider.
+//
+// Note that the provider interface is private as this is still experimental.
+// Use x/tracelog.New for an experimental trace log provider.
+func AddProvider(p provider) {
```
vk-outreach, line 41: Note that while this function is public, there is no real way to use this with anything other than pkg/x/tracelog.  Also, added this to avoid using a special trace.yaml config (based on feedback on the proposal).

pkg/trace/providers.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+
+	"github.com/getoutreach/gobox/internal/call"
+	"github.com/getoutreach/gobox/pkg/log"
+	"github.com/getoutreach/gobox/pkg/orerr"
+	"github.com/getoutreach/gobox/pkg/statuscodes"
+)
+
+// provider is a generic interface for tracking calls.
+type provider interface {
```
vk-outreach, line 14: This is only a subset of the interface proposed but felt like it was safer to start with the smallest step rather than going overboard here.

pkg/trace/call.go
```diff
@@ -101,25 +96,7 @@ func SetCallError(ctx context.Context, err error) error {
 // logging to happen (as do any SetCallStatus calls)
 func EndCall(ctx context.Context) {
 	defer End(ctx)
-
```
vk-outreach, line 104: This just got moved to providers.go as a default provider.

george-e-shaw-iv:
I just finished looking this over. Not approving it because its in draft form right now, but this looks great already.

jaredallard:
At the moment we don't really plan on reworking the existing logger, DT is going to budget redoing this in the future with some short-term hacks for costing. Closing this based on that, if you'd like to continue this PLEASE let us know and we can chat about the path forward!

PR #147, opened 2022-03-25
feat: Adding in closer() and rungroups to async

masterkidan:
Adds async functions for running a group of tasks and for closing out tasks.

## Jira ID

[QSS-836]

## Notes for your reviewers

[QSS-836]: https://outreach-io.atlassian.net/browse/QSS-836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/async/async.go
+++ b/pkg/async/async.go
@@ -136,8 +136,6 @@ func RunClose(ctx context.Context, r Runner) error {
 func RunGroup(rg []Runner) Runner {
 	ru := Func(func(ctx context.Context) error {
 		g, ctx := errgroup.WithContext(ctx)
-		ctx, cancelFunc := context.WithCancel(ctx)
-		defer cancelFunc()
 		for idx := range rg {
 			r := rg[idx]
 			g.Go(func() error {
@@ -147,11 +145,7 @@ func RunGroup(rg []Runner) Runner {
 					}
 				}()
 
-				err := r.Run(ctx)
-				if err != nil {
-					cancelFunc()
-				}
-				return err
+				return r.Run(ctx)
 			})
 		}
 		return g.Wait()
--- a/pkg/async/async.go
+++ b/pkg/async/async.go
@@ -32,9 +32,11 @@ package async
 import (
 	"context"
 	"errors"
+	"io"
 	"sync"
 	"time"
 
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
 
 	"github.com/getoutreach/gobox/pkg/events"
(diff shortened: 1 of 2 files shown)


pkg/async/async.go
```diff
@@ -114,6 +121,42 @@ func RunBackground(ctx context.Context, r Runner) {
...
+					if err := RunClose(ctx, r); err != nil {
+						log.Error(ctx, "Error when closing:", events.NewErrorInfo(err))
+					}
+				}()
+
+				err := r.Run(ctx)
+				if err != nil {
+					<-ctx.Done()
+				}
+				return err
```
vk-outreach, line 152: Why block here until context is canceled?  It will not allow an individual activity to error out and cancel the other activities?

```suggestion
				return r.Run(ctx)
```

masterkidan: So the g.wait will wait until so tasks are done before returning the first error. The behavior am trying to accomplish here is when the first routing errors or ask tasks exit and the error returns immediately

masterkidan: I was so focused on typing out the test I forgot what the actual code should be… one sec I’ll fix this up

masterkidan: hopefully this is better now

pkg/async/async.go
```diff
@@ -114,6 +121,44 @@ func RunBackground(ctx context.Context, r Runner) {
...
+			g.Go(func() error {
+				defer func() {
+					if err := RunClose(ctx, r); err != nil {
+						log.Error(ctx, "Error when closing:", events.NewErrorInfo(err))
+					}
+				}()
+
+				err := r.Run(ctx)
+				if err != nil {
+					cancelFunc()
```
vk-outreach, line 152: lol, this is exactly what `errgroup` itself does internally: see [here](https://cs.opensource.google/go/x/sync/+/036812b2:errgroup/errgroup.go;l=61) -- so you really do not need to do anything other than just `return r.Run(ctx)`

masterkidan: I guess I got confused by the comment on the wait method 
“ Wait blocks until all function calls from the Go method have returned, then
// returns the first non-nil error (if any) from them” 
Just assumed that meant it’s blocking till all the tasks finished

masterkidan: Learn something new everyday I guess. Thanks  for the help with this.

vk-outreach: Yeah, technically it does wait till all complete but it does cancel all the go-routines.  This is the same behavior in our current service activities I think

masterkidan: Yep it is. So maybe we could replace the service activities code with this now.... most of the pieces are here now ... with the closer & RunGroup, only thing left is to pass in the Config as part of the run method.

vk-outreach: Yeah we can switch to this once we change the service activities to take the `config` via the struct initialization rather than as argument to `Run`.

vk-outreach (approved):
LGTM

jaredallard (approved):

PR #151, opened 2022-04-11
feat(updater): add user configuration, markdown major release notes, spinner during check

jaredallard:

This PR adds some QoL adjustments to the updater framework, such as allowing users to opt-in to prereleases for all tools via the notion of `~/.outreach/.config/<repoName>/updater.yaml` (user config). Also added displaying major release notes when prompting -- much like Stencil does now, and a spinner because some updates can take awhile due to network latency and what not.

Also cleaned up a lot of the code to be easier to read / less complicated.

## Jira ID

[XX-XX]


bhogatemadhura:
is there a way to tell circle ci to use a pre-release version ?

pkg/updater/github.go
```diff
@@ -101,16 +103,13 @@ func (g *Github) GetLatestVersion(ctx context.Context, currentVersion string, in
...
-			version = parsedVersion
-		}
+	version, err := semver.ParseTolerant(currentVersion)
+	if err != nil {
+		// if we can't determine the version, fallback to empty (oldest) version
+		version = semver.MustParse("0.0.0")
 	}
 
-	// Skip pre versions that don't aren't rc (made by bootstrap)
+	// Skip pre versions that aren't rc (made by bootstrap)
```
plisy, line 112: Why though?

pkg/updater/github.go
```diff
@@ -303,37 +301,39 @@ func (g *Github) DownloadRelease(ctx context.Context, r *github.RepositoryReleas
...
 // as the current executable. If it is not an archive then the file given is returned
 func (g *Github) processArchive(ctx context.Context, file, tmpDir, execName string, asset *github.ReleaseAsset) (string, error) {
-	// TODO: better support for other archives in the future
-	if strings.HasSuffix(asset.GetName(), ".tar.gz") {
-		f, err := os.Open(file)
-		if err != nil {
-			return "", errors.Wrap(err, "failed to open downloaded archive")
-		}
-		defer f.Close()
+	if !strings.Contains(asset.GetName(), ".tar") {
```
plisy, line 311: Nit: doing `!strings.HasSuffix(asset.GetName(), ".tar.gz") && !strings.HasSuffix(asset.GetName(), ".tar.xz")` seems safer.

pkg/updater/github.go
```diff
@@ -303,37 +301,39 @@ func (g *Github) DownloadRelease(ctx context.Context, r *github.RepositoryReleas
...
+		return "", errors.Wrap(err, "failed to open downloaded archive")
+	}
+	defer f.Close()
 
-		// use the name of the executable here to allow for multiple clis in a given repository
-		file, err = g.getFileFromArchive(ctx, f, storageDir, execName)
-		if err != nil {
-			return "", errors.Wrap(err, "failed to extract archive")
-		}
+	storageDir := filepath.Join(tmpDir, strings.TrimSuffix(asset.GetName(), ".tar.gz"))
```
plisy, line 321: You check for `.tar` but expect `.tar.gz`

plisy: If there's a reason, it needs to be a comment.

pkg/updater/github.go
```diff
@@ -357,57 +357,74 @@ func (g *Github) ReplaceRunning(ctx context.Context, newBinary string) error {
...
 	}
 
 	return file, nil
 }
 
-// finds specified file in provided tar archive and returns io.Reader and its size for extraction progress
-func findTarFile(ctx context.Context, archive *tar.Reader, filename string) (io.Reader, int64, error) {
+// findFileInTarFile finds a specified file in provided tar archive and returns io.Reader
+// and its size for extraction progress
+func findFileInTarFile(ctx context.Context, archive *tar.Reader, filename string) (io.Reader, int64, error) {
```
plisy, line 427: How does this handle subdirs?

pkg/updater/updater.go
```diff
@@ -88,22 +83,28 @@ func NeedsUpdate(ctx context.Context, log logrus.FieldLogger, repo, version stri
 		log.WithError(err).Warn("failed to get user's home directory")
 		return false
 	}
+	cacheDir := filepath.Join(homedir, ".outreach", ".cache", ".updater")
```
plisy, line 86: Should this follow the same app/repo pattern config does? (Consistency, feel free to ignore)

bhogatemadhura (approved):

bhogatemadhura:
> > is there a way to tell circle ci to use a pre-release version ?
> 
> Yes, when downloading the release binary. Or, you could just run it with `--force-update --enable-prereleases`, that behaviour hasn't changed.

not entirely sure about the process yet. can you elaborate a bit please ?
I meant .. if I open a PR and circle CI runs e2e test, how can I tell it to run pre-release version of devenv ? Is that possible ?

pkg/updater/github.go
jaredallard: Eh it's old logic that has since lost reasoning because of a lack of a comment.

pkg/updater/github.go
jaredallard: Good catch, I just need to remove the extension here

pkg/updater/github.go
jaredallard: It's like file.Watch, it doesn't need to handle them it'll just be walked over recursively

pkg/updater/updater.go
jaredallard: It does, but it uses repo.yaml instead of a dedicated directory

pkg/updater/github.go
jaredallard: I wanted to have less statements, and then error in the handler later on so this was intentional. The actual decompression function handles unknown formats

jaredallard:
> > > is there a way to tell circle ci to use a pre-release version ?
> 
> > 
> 
> > Yes, when downloading the release binary. Or, you could just run it with `--force-update --enable-prereleases`, that behaviour hasn't changed.
> 
> 
> 
> not entirely sure about the process yet. can you elaborate a bit please ?
> 
> I meant .. if I open a PR and circle CI runs e2e test, how can I tell it to run pre-release version of devenv ? Is that possible ?

That is currently not possible, no. It's also out of scope of this work, which was just to enhance the updater framework to support setting global prerelease usage. Binaries aren't meant to be updated at all in CI and frequently skip this altogether

plisy (approved):

jaredallard:
ok idk what's up w/ coveralls but I tried my best

PR #152, opened 2022-04-18
feat(serviceactivity): implement serviceactivity framework

jaredallard:

This PR implements a serviceactivity package for replacing the current service activity framework in Bootstrap for the future. A notable change is the removal of config from the `Run` function on the `ServiceActivity` interface. We argue that that should be passed in the constructor of a service activity, not in the actual run call itself. This creates a breaking change, which would be rolled out in Q2 with the other breaking changes in `main.go`.

## Jira ID

[XX-XX]

### DIFF
--- a/pkg/serviceactivity/runner.go
+++ b/pkg/serviceactivity/runner.go
@@ -1,0 +1,40 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
+
+// Description: This file implements the service activity runner
+
+package serviceactivity
+
+import (
+	"context"
+
+	"golang.org/x/sync/errgroup"
+)
+
+// Runner is a service activity runner
+type Runner struct {
+	acts []ServiceActivity
+}
+
+// New creates a new service activity runner
+func New(acts []ServiceActivity) *Runner {
+	return &Runner{acts}
+}
+
+// Run starts all serviceactivities and blocks until they are done
+// returning an error if any of them failed to start.
+func (r *Runner) Run(ctx context.Context) error {
+	g, ctx := errgroup.WithContext(ctx)
+
+	// run the activities in parallel
+	for idx := range r.acts {
+		act := r.acts[idx]
+		g.Go(func() error {
+			// stop the activity once it returns
+			defer act.Close(ctx)
+
+			return act.Run(ctx)
+		})
+	}
+
+	return g.Wait()
+}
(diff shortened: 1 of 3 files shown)


pkg/serviceactivity/serviceactivity.go
```diff
@@ -0,0 +1,18 @@
...
+// Package serviceactivity provides service activities for services
+// to implement and a framework to run them.
+package serviceactivity
+
+import (
+	"context"
+)
+
+// ServiceActivity is the interface that all service activities must implement.
+type ServiceActivity interface {
```
george-e-shaw-iv, line 15: `serviceactivity.ServiceActivity` come on

pkg/serviceactivity/serviceactivity.go
```diff
@@ -0,0 +1,18 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
+
+// Description: This file defines the ServiceActivity type
+// to be implemented by services.
+
+// Package serviceactivity provides service activities for services
+// to implement and a framework to run them.
+package serviceactivity
```
george-e-shaw-iv, line 8:
```suggestion
package service
```

pkg/serviceactivity/serviceactivity.go
george-e-shaw-iv, line 15:
```suggestion
type Activity interface {
```

pkg/serviceactivity/runner.go
```diff
@@ -0,0 +1,40 @@
...
+// returning an error if any of them failed to start.
+func (r *Runner) Run(ctx context.Context) error {
+	g, ctx := errgroup.WithContext(ctx)
+
+	// run the activities in parallel
+	for idx := range r.acts {
+		act := r.acts[idx]
+		g.Go(func() error {
+			// stop the activity once it returns
+			defer act.Close(ctx)
```
george-e-shaw-iv, line 33: close needs a new context

pkg/serviceactivity/serviceactivity.go
jaredallard: 🙉 ❤️

pkg/serviceactivity/serviceactivity.go
jaredallard: but it's not a generic service package 😉

pkg/serviceactivity/runner.go
jaredallard: that's been a bug forever, wow

pkg/serviceactivity/serviceactivity.go
george-e-shaw-iv: could it ever be? i still think my naming suggestions are better regardless

jaredallard:
Spoke with VK, we're going to move this to be a `RunCloser` in async instead. Will open a new PR with these changes and back link.

PR #157, opened 2022-04-27
feat: Added basic Telefok client usable as trace presend hook

plisy:

Adding CLI telemetry for bootstrap CLIs. It goes through telefork as that's completely public with no real secrets, so it can go OSS.

Also

image

## Jira ID

[DTSS-1629]

## Notes for your reviewers

[DTSS-1629]: https://outreach-io.atlassian.net/browse/DTSS-1629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/cli/bootstrap.go
+++ b/pkg/cli/bootstrap.go
@@ -106,7 +106,7 @@ func urfaveRegisterShutdownHandler(cancel context.CancelFunc) {
 	}()
 }
 
-func addCommonProps(ctx context.Context, c telefork.Client) {
+func commonProps() log.Marshaler {
 	commonProps := log.F{
 		"os.name": runtime.GOOS,
 		"os.arch": runtime.GOARCH,
@@ -132,8 +132,8 @@ func addCommonProps(ctx context.Context, c telefork.Client) {
 			}
 		}
 	}
-	trace.AddInfo(ctx, commonProps)
-	c.AddInfo(commonProps)
+
+	return commonProps
 }
 
 // setupTracer sets up a root trace for the CLI and initializes the tracer
@@ -193,9 +193,11 @@ func HookInUrfaveCLI(ctx context.Context, cancel context.CancelFunc, a *cli.App,
 		c.SendEvent(e)
 	})
 
-	addCommonProps(ctx, c)
+	props := commonProps()
+	c.AddInfo(props)
 
 	ctx = setupTracer(ctx, a.Name)
+	trace.AddInfo(ctx, props)
 
 	exitCode, exit, cleanup := setupExitHandler(ctx)
 	defer func() {
--- a/pkg/cli/bootstrap.go
+++ b/pkg/cli/bootstrap.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	osexec "os/exec"
 	"os/signal"
 	"os/user"
 	"runtime"
@@ -22,6 +23,7 @@ import (
 	"github.com/getoutreach/gobox/pkg/exec"
 	"github.com/getoutreach/gobox/pkg/log"
 	"github.com/getoutreach/gobox/pkg/secrets"
+	"github.com/getoutreach/gobox/pkg/telefork"
 	"github.com/getoutreach/gobox/pkg/trace"
 	"github.com/getoutreach/gobox/pkg/updater"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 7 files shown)


george-e-shaw-iv (approved):

pkg/telefork/client.go
```diff
@@ -0,0 +1,110 @@
+package telefork
```
ChrisPRobinson, line 1: Can we add even a really basic test here? Just a quicker drive by one? I feel like it sets things up better in case more changes are made in the future.

pkg/cli/bootstrap.go
```diff
@@ -104,6 +106,36 @@ func urfaveRegisterShutdownHandler(cancel context.CancelFunc) {
 	}()
 }
 
+func addCommonProps(ctx context.Context, c telefork.Client) {
```
ChrisPRobinson, line 109: This function does 3 things, builds the common props, adds it to trace and adds it to the client. Should it just do one thing instead, build the CommonProps?

pkg/telefork/client.go
```diff
@@ -0,0 +1,110 @@
...
+type client struct {
+	appName  string
+	apiKey   string
+	endpoint string
+	events   []Event
+
+	commonProps map[string]interface{}
+}
+
+func (c *client) SendEvent(event Event) {
```
ChrisPRobinson, line 45: Reading the code it confuses me that we have a method "SendEvent" that simply appends and "Close" that actually sends.

ChrisPRobinson: Can we make it "AppendEvent" and "Flush" instead?

pkg/telefork/client.go
```diff
@@ -0,0 +1,110 @@
...
+	delete(e, "meta.beeline_version")
+	delete(e, "meta.local_hostname")
+	delete(e, "meta.span_type")
+
+	c.events = append(c.events, e)
+}
+
+func (c *client) Close() {
+	if c.apiKey == "" || c.apiKey == "NOTSET" {
+		return
```
ChrisPRobinson, line 67: If things are not configured right then nothing will be sent and there is no error message, should we add a printlin that is a warning that this has occurred?

pkg/telefork/telefork.go
```diff
@@ -0,0 +1,5 @@
+// Package telefork contains basic telefork client and exposes tracing presend handler.
+//
+// Package telefork contains basic telefork client and exposes tracing presend handler.
+// This package is used in bootstrap CLIs for basic telemetry.
+package telefork
```
ChrisPRobinson, line 5: Its too bad this package can't just go in telefork, but it's not a public package and this is
.

ChrisPRobinson (commented):
Noticed a couple things, though understand were trying to move quick here.

pkg/telefork/client.go
plisy: Done

pkg/telefork/client.go
plisy: This will kinda change over time. The Flush on Close is the simplest implementation. Eventually I'd like to add some reliability.

ChrisPRobinson (approved):
Yay to tests! My other comments are not big things, would rather not delay and get things in, with tests we can always refactor easier later, KISS makes sense at this stage

pkg/telefork/client.go
plisy: And flush periodically not just on close.

pkg/telefork/client.go
plisy: I'm not sure this shouldn't just stay silent, telemetry is kinda optional (also, this is set during build, so if it's not, then that's build error not a runtime error)

bhogatemadhura (approved):

pkg/telefork/telefork.go
plisy: I agree. If I was going to spend more time on this now, I'd probably make it a generic HTTP thing instead, and configure it in bootstrap. That way there would be a good separation between things.

That said, it seems too much work at this point. I'd rather get more detailed telemetry out of CLIs, if I can choose.

pkg/telefork/client.go
ChrisPRobinson: Understood, just reading the code on first blush was confusing though KISS makes sense

pkg/telefork/client.go
ChrisPRobinson: Ok sounds good.

pkg/telefork/telefork.go
ChrisPRobinson: Totally agree

pkg/telefork/client.go
plisy: I'll need a way to do telemetry across bin runs (one exec starts span, other exec terminates span). And there reliability will matter way more.

PR #161, opened 2022-05-04
fix: rebootstrap library

george-e-shaw-iv:
## Jira ID

[XX-XX]


george-e-shaw-iv (commented):
lgtm

service.yaml
```diff
@@ -29,10 +33,20 @@ arguments:
     allowMajorVersions: true
     enablePrereleases: false
   reportingTeam: fnd-dtss
-  type:
-    - library
+  service: false
+  tutorial: false
+  vaultSecrets:
```
plisy, line 38: Should this be leaked in public repo?

george-e-shaw-iv: they're just paths and jared needed them there to make something work. so idk

plisy (approved):

PR #167, opened 2022-05-19
feat(trace): move to opentelemetry tracing

ghost:

This moves us to using OpenTelemetry for tracing while still supporting Honeycomb based propagation from other services.

All the capabilities previously present should still function with some modifications:

- Removes `ContextFromHTTP` and replaced the functionality with `NewHandler` and `NewTransport`
- `ForceTracing` must now get called before any traces start and applies to all traces

The below examples show `stonehenge` migrated to OpenTelemetry tracing. The example receives a Honeycomb backed GRPC call for `giraffe` and calls a Honeycomb backed HTTP endpoint in the `flagship`. You can see that things look a little different with OpenTelemetry, but all the important information seems to be there.

- Honeycomb (old): https://ui.honeycomb.io/outreach-a0/datasets/dev/result/JvttbaJuL3y/trace/sgBLRFHUCrJ?fields[]=name&fields[]=service_name&span=1d1201b31030a2b0
- Otel (new): https://ui.honeycomb.io/outreach-a0/datasets/dev/result/g3UgbuPdGiK/trace/3inTj8CyFQt?fields[]=name&fields[]=service_name&span=7925efc3b34430da

## Jira ID

[DTSS-1737]

## Notes for your reviewers

I have opted for removing the `ContextFromHTTP` interface for two reasons:

1. Supporting both Honeycomb and OTel headers simultaneously required some additional wrapping that was easier done with all the information than just the headers. The automatic instrumentations handles things like baggage which would add significant complexity to the `ContextFromHTTP` paradigm. 
2. OTel introduces some really nice instrumentation that we can now take advantage of.

I kept `FromHeader` and `ToHeader` for `grpcx` which relies on these for multiple things. Ideally, we can move away from these in the future and replace the functionality with OTel instrumentation. This would benefit this in the same way as mentioned above. The move will be much easier once we can drop support for Honeycomb headers.

[DTSS-1737]: https://outreach-io.atlassian.net/browse/DTSS-1737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/trace/propagation.go
+++ b/pkg/trace/propagation.go
@@ -3,6 +3,7 @@ package trace
 import (
 	"context"
 	"net/http"
+	"strings"
 
 	"github.com/honeycombio/beeline-go/propagation"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
@@ -15,6 +16,8 @@ import (
 const (
 	// Header that enforces the tracing for particular request
 	HeaderForceTracing = "X-Force-Trace"
+	// Header used by OpenTelemetry to propagate traces
+	OtelPropagationHeader = "Traceparent"
 )
 
 type roundtripper struct {
@@ -23,7 +26,7 @@ type roundtripper struct {
 
 func (rt roundtripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	headers := map[string]string{
-		"traceparent": r.Header.Get("Traceparent"),
+		strings.ToLower(OtelPropagationHeader): r.Header.Get(OtelPropagationHeader),
 	}
 	_, prop, err := propagation.UnmarshalW3CTraceContext(r.Context(), headers)
 	if err != nil {
@@ -54,7 +57,7 @@ type Handler struct {
 }
 
 func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if r.Header.Get("Traceparent") == "" {
+	if r.Header.Get(OtelPropagationHeader) == "" {
 		hcHeaderToW3CHeaders(r)
 	}
 
@@ -100,8 +103,9 @@ func (t *otelTracer) toHeaders(ctx context.Context) map[string][]string {
 		propagator := otel.GetTextMapPropagator()
 		propagator.Inject(ctx, otelpropagation.HeaderCarrier(result))
 
+		// Honeycomb expects
 		headers := map[string]string{
-			"traceparent": result.Get("Traceparent"),
+			strings.ToLower(OtelPropagationHeader): result.Get(OtelPropagationHeader),
 		}
 
 		_, prop, err := propagation.UnmarshalW3CTraceContext(ctx, headers)
@@ -126,7 +130,7 @@ func (t *otelTracer) fromHeaders(ctx context.Context, hdrs map[string][]string, 
 	}
 
 	if defaultTracer != nil {
-		if header.Get("Traceparent") == "" {
+		if header.Get(OtelPropagationHeader) == "" {
 			prop, err := propagation.UnmarshalHoneycombTraceContext(header.Get(propagation.TracePropagationHTTPHeader))
 			if err == nil {
 				_, headers := propagation.MarshalW3CTraceContext(ctx, prop)


pkg/trace/otel.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// Annotator is a SpanProcessor that adds service-level tags on every span
+type Annotator struct {
+	globalTags GlobalTags
+}
+
+func (a Annotator) OnStart(_ context.Context, s sdktrace.ReadWriteSpan) {
+	setf := func(key string, value interface{}) {
+		s.SetAttributes(attribute.String(key, fmt.Sprintf("%v", value)))
```
richardlarocque, line 38: I have some concerns about the `%v` formatter here.  I think there are cases where the input contains one or more nested maps.  

See https://github.com/getoutreach/gobox/blob/v1.40.0/pkg/log/logtest/logtest.go#L70 and https://github.com/getoutreach/gobox/blob/c8a5b56788ef7a8ad27373c63febf6f993e1a1b0/internal/logf/logf.go#L37 for some tools that might help to convert maps to something dot-separated.

richardlarocque: I'll grant that this seems to be equally unhandled in the Honeycomb implementation.

I don't know if that means this has been broken forever, or if I'm currently misinterpreting how it works.

pkg/trace/otel.go
```diff
@@ -0,0 +1,195 @@
...
+	)
+	if err != nil {
+		log.Error(ctx, "Unable to configure trace provider", events.NewErrorInfo(err))
+	}
+
+	tp := sdktrace.NewTracerProvider(
+		sdktrace.WithBatcher(exp),
+		sdktrace.WithResource(r),
+		// otel accepts sample rates as a fractions of traces >= 1 will always sample < 0 will never sample
+		sdktrace.WithSampler(sdktrace.TraceIDRatioBased(t.Otel.SamplePercent/100)),
```
richardlarocque, line 104: Do you know if this uses the same sampling logic as the original Honeycomb sampler?

We need to be careful about changing this logic, since we rely on different services to come to the same conclusions about the same trace IDs.  If they're different, we'll get traces from some services but not others.

pkg/trace/otel.go
```diff
@@ -0,0 +1,195 @@
...
+	if span := trace.SpanFromContext(ctx); span != nil {
+		span.End()
+	}
+}
+
+func (t *otelTracer) addInfo(ctx context.Context, args ...log.Marshaler) {
+	if span := trace.SpanFromContext(ctx); span != nil {
+		for _, f := range args {
+			logf.Marshal("", f, func(key string, value interface{}) {
+				span.SetAttributes(attribute.String(key, fmt.Sprintf("%v", value)))
```
richardlarocque, line 180: Again, the `%v` formatter seems suspect, see above.

pkg/trace/otel.go
```diff
@@ -0,0 +1,195 @@
...
+
+func (t *otelTracer) spanID(ctx context.Context) string {
+	if span := trace.SpanFromContext(ctx); span != nil && span.SpanContext().SpanID().IsValid() {
+		return span.SpanContext().SpanID().String()
+	}
+	return ""
+}
+
+func (t *otelTracer) parentID(ctx context.Context) string {
+	return ""
```
richardlarocque, line 194: This doesn't look right.  If this is intentional, maybe add a comment?

pkg/trace/propagation.go
```diff
@@ -30,14 +46,26 @@ func (t *tracer) fromHeaders(ctx context.Context, hdrs map[string][]string, name
...
+
+// fromHeaders fetches trace info from a headers map
+func (t *otelTracer) fromHeaders(ctx context.Context, hdrs map[string][]string, name string) context.Context {
+	return ctx
+}
+
+// toHeaders writes the current trace context into a headers map
+func (t *otelTracer) toHeaders(ctx context.Context) map[string][]string {
+	return http.Header{}
+}
```
richardlarocque, line 71: I think I know why these are blank (this is handled automatically by the round-tripper, right?), but it might help to add a comment about it.

richardlarocque (commented):
Sorry this took so long to review.  It fell off my to-do list for a bit.

This first round of reviews is looking just at non-test code.  I figured I should make sure I understand the code before diving into tests.

richardlarocque:
OpenTelemetry brings some newly standardized field names.  It seems like this PR is trying to keep with the existing practice where we mostly set our own field names, according to [our own standards](https://outreach-io.atlassian.net/wiki/spaces/EN/pages/691405109/Logging+Standards).  

I presume we'll want to move to OpenTelemetry standards at some point.  That would bring a lot of benefit, especially in projects in less-supported languages (Python, Elixir, Node.js) where we could really use good auto-instrumentation.  But I'm also going to guess that we don't want to go that route right now, since it will be disruptive to existing practice.  In other words, we're trying to change as little as possible, for now.

Is there any worry that fields that were automatically set by the Honeycomb tracer could no longer be set by the OpenTelemetry tracer?  Do you have a sense as to whether or not we do rely on Honeycomb-specific names or behaviors that don't carry over to OpenTelemetry?

pkg/trace/propagation.go
artemchernyak-outreach: This was an unfortunate byproduct of putting OTel and Honeycomb under the same interface.

As you mentioned in another comment, OTel introduces really great support for auto instrumentation that would take care of 80% of what we generate by hand. This instrumentation also handles propagation for us and completely gets rid of the need for manual header handling.

I am going to look at moving towards that approach in another PR. My idea is to replace the manual handling, for both OTel and Honeycomb, with a wrapper similar to the automatic OTel instrumentation. You can see an example of the automatic instrumentation here https://github.com/getoutreach/gobox/pull/167/files#diff-d36e35946980b42d4d9b45ea6567e51f1efb2d63bb5078c89abc2b7f83c36e4aR181.

pkg/trace/otel.go
artemchernyak-outreach: Updated with deprecation comments.

pkg/trace/honeycomb.go
```diff
@@ -13,6 +14,11 @@ import (
 	"github.com/honeycombio/beeline-go/trace"
 )
 
+type honeycombTracer struct {
+	Config
+	sync.Once
```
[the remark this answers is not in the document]
artemchernyak-outreach: Good call! Doesn't look like it was being used anywhere. Removed.

pkg/trace/otel.go
richardlarocque: I was kind of hoping they'd use the same algorithm.  My read is that they're not the same.

- https://github.com/honeycombio/beeline-go/blob/v1.8.0/sample/deterministic_sampler.go
- https://github.com/open-telemetry/opentelemetry-go/blob/sdk/v1.7.0/sdk/trace/sampling.go#L82-L99

They have a similar flavor to them, but the opentelemetry one is using 8 bytes of the trace ID, while the beeline-go one is using 4 bytes.

If I'm right about this, then a service in "opentelemetry mode" will chose to sample a different set of traces than applications in "honeycomb mode".  That's going to make it hard to reconstruct full traces across all our services.

richardlarocque (approved):
I skimmed through the tests; they look mostly fine.

I do have some remaining concerns:
- If I'm right about the different sampling algorithms, we could see broken traces in production as we migrate over to this system.
- Serializing span values with `%v` is not great.  I can't argue about it too much if that's what Honeycomb already does, though.

These are relatively minor compared to the overall size of what you're trying to do here, though.

I'm going to hit the "approve" button in case you want to move forward with this as-is.  All the new stuff is behind a config, and I don't see anything in here that's likely to break existing functionality.

If you want to wait for further reviews from VK or DTSS, that's also fine.  I bet they'd have useful comments, too.

pkg/trace/force_sampler.go
```diff
@@ -57,3 +58,53 @@ func sampleAt(ctx context.Context, rate uint) context.Context {
...
+//nolint:gocritic // Why: Required to pass SamplingParameters as a copy
+func (s *otelForceSampler) ShouldSample(p sdktrace.SamplingParameters) sdktrace.SamplingResult {
+	psc := oteltrace.SpanContextFromContext(p.ParentContext)
+	sampler, err := sample.NewDeterministicSampler(s.sampleRate)
+	if err != nil {
+		panic(err)
+	}
+
+	var forceTrace bool
+	for _, a := range p.Attributes {
```
richardlarocque, line 79: I'll probably regret this question, but out of curiosity:

Why do we need to cast `a.Key` to a `string`?  Would it be possible to do something like `p.Attributes[fieldForceTrace]` and skip iterating through all the keys like this?  Is `p.Attributes` not a simple `map[string]Value`?

pkg/trace/force_sampler.go
```diff
@@ -57,3 +58,53 @@ func sampleAt(ctx context.Context, rate uint) context.Context {
...
+		if string(a.Key) == fieldForceTrace {
+			forceTrace = a.Value.AsBool()
+		}
+	}
+
+	if forceTrace {
+		return sdktrace.SamplingResult{
+			Decision:   sdktrace.RecordAndSample,
+			Tracestate: psc.TraceState(),
+		}
```
richardlarocque, line 89: Honeycomb is pretty smart about sampling.

It knows what kind of dice we're rolling and can adjust counts accordingly.  If your services is sampling 1/100, it'll note this in the trace data, and in the UI it'll show all counts multiplied by 100 to make up for it.  If you decide to adjust your sampling rate to 1/200 the next day, it'll keep adjusting and the graphs in the UI will continue to look mostly the same.

If you have a *forced* trace, that means you have a sample rate of 100% for that particular trace.  Honeycomb knows how to account for that, too.  A forced trace will increment the counts by exactly one in the UI.  That way a couple of forced traces won't cause your graphs to look weird.

To support all this, Honeycomb samplers return not just a true/false, but also the sample rate used for that trace.  See [here](https://github.com/getoutreach/outreach-ruby-lib/blob/80dd73b3d1732cb363a963392ce3d07503714582/lib/outreach_trace/allow_force_sampler.rb#L27-L35).

Which brings me to the point of this long comment.  We don't seem to be returning anything about sampling rates here.  Does OTel not support any kind of concept like this?

pkg/trace/otel.go
```diff
@@ -199,3 +200,11 @@ func (t *otelTracer) spanID(ctx context.Context) string {
 func (t *otelTracer) parentID(ctx context.Context) string {
 	return ""
 }
+
+func (t *otelTracer) setForce(force bool) {
+	t.force = force
+}
```
richardlarocque, line 206: In at least *some* of the tracing implementations, I tried to make "force-trace" a string rather than a boolean, then include that string value as a trace tag.

That way, I can do `curl -H 'X-Force-Trace: rltesthackstuff' ...` then go look for traces in the Honeycomb UI that have "force_trace == rltesthackstuff".  

(Admittedly, I don't think I've been consistent about this, and some services do still flatten this to a bool.  I haven't gotten around to fixing it all yet.  Still, when it works, it's pretty great.)

richardlarocque (commented):
Mostly good, I think.  I don't have a very strong understanding of the `otel` APIs, so kind of glossing over those parts.

pkg/trace/force_sampler.go
artemchernyak-outreach: Unfortunately it's not. And I really don't like this either.

`p.Attributes` is actually a `[]attribute.KeyValue` which has `Key` and `Value` types with their own wrappers for Go primitive types like `String`, `Bool`, etc.

That makes the api pretty ugly to work with when we need to pull off some values.

pkg/trace/force_sampler.go
artemchernyak-outreach: It looks like honeycomb supports this in OpenTelemetry through the use of attributes. Their documentation is a little confusing about how to accomplish this, but looks like it is possible. I will have to dig into it.

pkg/trace/otel.go
artemchernyak-outreach: Oh I like that idea! I'll update things to support this.

pkg/trace/force_sampler.go
artemchernyak-outreach: Added the new attribute in accordance with Honeycomb's documentation https://docs.honeycomb.io/getting-data-in/opentelemetry/go/#sampling

pkg/trace/otel.go
artemchernyak-outreach: Stored as a string and forces on any non empty string.

artemchernyak-outreach:
I am going to hold off on merging this until I make corresponding updates to httpx and grpcx. That should give us a chance to actually test this with our existing services and iron out any issues that might arise there.

pkg/app/info.go
```diff
@@ -162,6 +162,7 @@ type Data struct {
 func (d *Data) MarshalLog(addField func(key string, v interface{})) {
 	if d.Name != "unknown" {
 		addField("app.name", d.Name)
+		addField("service_name", d.Name)
```
bhogatemadhura, line 165: should it be `service.name` ?

ChrisPRobinson (approved):
LGTM

pkg/trace/call_test.go
```diff
@@ -39,142 +39,113 @@ func (suite) TestNestedCall(t *testing.T) {
 	defer app.SetName(app.Info().Name)
 	app.SetName("gobox")
 
-	trlog := tracetest.NewTraceLog()
```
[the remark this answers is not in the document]
artemchernyak-outreach: This actually wasn't removed. Just the order has changed. This is now after the `expected` declaration. I do agree that the diff makes this look really confusing.

pkg/trace/call_test.go
```diff
@@ -39,142 +39,113 @@ func (suite) TestNestedCall(t *testing.T) {
...
+			"attributes.error.kind":          "error",
+			"attributes.error.error":         "sql error",
+			"attributes.error.message":       "sql error",
+			"attributes.error.stack":         differs.AnyString(),
+			"attributes.info_1_key":          "info_1_val",
+			"attributes.info_2_key":          "info_2_val",
+			"attributes.timing.dequeued_at":  differs.RFC3339NanoTime(),
+			"attributes.timing.finished_at":  differs.RFC3339NanoTime(),
+			"attributes.timing.scheduled_at": differs.RFC3339NanoTime(),
+			"attributes.timing.service_time": differs.AnyString(),
```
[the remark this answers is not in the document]
artemchernyak-outreach: Yeah, this is something Richard mentioned too. So OTel attributes require you to specify the type. Honeycomb attributes took an `interface{}` and handled that correctly. The way we add attributes using a logger passes an `interface{}` everywhere. 

How the logger works left me with two options: cast everything to the same type, I chose string since I can just use `fmt.Sprinf`, or do some reflection magic to determine the type. As this PR was already getting complex, I went for the simpler solution. But I am happy to update it in a follow up to use reflection and set the right types.

From the Honeycomb perspective, it looks like Honeycomb was smart to automatically convert this string.

pkg/app/info.go
kadukf-outreach: @artemchernyak-outreach +1 for consistency + ticket for deprecating app.name ?

samnguyen-outreach:
FYI there is a breaking change in here that I had to dig around to find out. I had to make minor adjustments but you might want to make a wider communication (at least I couldn't find any with my searching abilities)

These were the changes I had to make in my code:
```diff
-       traceLog := tracetest.NewTraceLog()
+       spanRecorder := tracetest.NewSpanRecorder()

-       assert.Assert(t, len(traceLog.HoneycombEvents()) > 0,
+       spans := spanRecorder.Ended()

```

artemchernyak-outreach:
Thanks for the note @samnguyen-outreach. I'll add this information to the original announcement, that way it will be easy to find in the future.

PR #168, opened 2022-05-21
fix: constant updating

ChrisPRobinson:
fix: enable `skip-update` to work right

Previously the string was the wrong name so it never skipped.

## Jira ID

[DTSS-1841]

## Notes for your reviewers

[DTSS-1841]: https://outreach-io.atlassian.net/browse/DTSS-1841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/updater/cli.go
+++ b/pkg/updater/cli.go
@@ -56,7 +56,7 @@ func (u *updater) hookIntoCLI() {
 				u.forceCheck = true
 			}
 
-			if strings.EqualFold(f, "skip-updater") {
+			if strings.EqualFold(f, "skip-update") {
 				u.disabled = true
 			}
 		}


plisy (approved):

pkg/updater/cli.go
```diff
@@ -56,7 +56,7 @@ func (u *updater) hookIntoCLI() {
 				u.forceCheck = true
 			}
 
-			if strings.EqualFold(f, "skip-updater") {
+			if strings.EqualFold(f, "skip-update") {
```
jaredallard, line 59: Whew

PR #180, opened 2022-06-23
feat: force no trace

george-e-shaw-iv:

* adds the ability to force no trace from headers

## Jira ID

[XX-XX]

### DIFF
--- a/pkg/trace/propagation.go
+++ b/pkg/trace/propagation.go
@@ -69,10 +69,15 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	handler := otelhttp.NewHandler(h.handler, h.operation)
 
-	force := r.Header.Get(HeaderForceTracing)
-	if force != "" {
-		startOptions = oteltrace.WithAttributes(attribute.String(fieldForceTrace, force))
+	forceTrace := r.Header.Get(HeaderForceTracing)
+	if forceTrace != "" {
+		startOptions = oteltrace.WithAttributes(attribute.String(fieldForceTrace, forceTrace))
 		handler = otelhttp.NewHandler(h.handler, h.operation, otelhttp.WithSpanOptions(startOptions))
+	} else {
+		forceNoTrace := r.Header.Get(HeaderForceNoTracing)
+		if forceNoTrace != "" {
+			handler = h.handler
+		}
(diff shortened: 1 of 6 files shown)


pkg/trace/propagation.go
```diff
@@ -127,6 +131,12 @@ func (t *otelTracer) fromHeaders(ctx context.Context, hdrs map[string][]string,
 	force := header.Get(HeaderForceTracing)
 	if force != "" {
 		ctx = ForceTracing(ctx)
+	} else {
+		// We do this in an else block so force-trace always takes precedent over force-no-trace.
+		forceNoTracing := header.Get(HeaderForceNoTracing)
+		if forceNoTracing != "" {
+			ctx = ForceNoTracing(ctx)
```
artemchernyak-outreach, line 138: This should just skip creating the `otelhttp.NewHandler` and just run `h.ServeHTTP(w, r)` here. Otherwise `otelhttp.NewHandler` will continue propagating the handler.

artemchernyak-outreach (changes requested):

pkg/trace/propagation_test.go
```diff
@@ -14,6 +14,40 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
+func (suite) TestForceNoTracingByHeader(t *testing.T) {
+	defer app.SetName(app.Info().Name)
+	app.SetName("gobox")
+
+	recorder := tracetest.NewSpanRecorderWithOptions(tracetest.Options{
+		SamplePercent: 0.1,
```
artemchernyak-outreach, line 22: You should use `NewSpanRecorder` here without the options. That way you will use 100% sampling instead of 0.1%.

pkg/trace/propagation.go
artemchernyak-outreach: If you don't want to completely drop the trace, then you can manually extract it like https://github.com/getoutreach/gobox/blob/main/pkg/trace/propagation.go#L140-L141. Just don't start a new span. 

I could see this header working either way. But to me it seems more intuitive to have this completely drop the trace is you pass a no tracing header.

george-e-shaw-iv: i think you thought this was in another function, and i added it in the appropriate function

PR #185, opened 2022-07-13
fix(async.pool): making logging optional

kadukf-outreach:
Logging in the async.pool is noisy, errors can be a business values handled in the business code (moving/extracting the code)

## Jira ID

[XX-XX]


pkg/async/pool/pool.go
```diff
@@ -273,3 +262,27 @@ type unit struct {
 	Context context.Context
 	Runner  async.Runner
 }
+
+type loggingScheduler struct {
+	Inner Scheduler
+	Name  string
```
dmytrokasianenko-outreach, line 268: there is already option for Name, would it make sense to remove it?

kadukf-outreach: removing it

dmytrokasianenko-outreach (approved):

pkg/async/pool/pool.go
```diff
@@ -273,3 +255,27 @@ type unit struct {
...
+}
+
+func (w *loggingScheduler) Schedule(ctx context.Context, r async.Runner) error {
+	return w.log(ctx, w.Inner.Schedule(ctx, async.Func(func(ctx context.Context) error {
+		return w.log(ctx, r.Run(ctx))
+	})))
+}
+
+// WithLogging creates a scheduler which logs the errors returned from the scheduling as well as executing phase
```
artemchernyak-outreach, line 270: Does this mean that if a service update gobox without switching over to `WithLogging` they will lose logging without warning?

If that's the case, I think that we should make this `WithoutLogging` and make the default include logging.

vk-outreach: Yeah, looks like this is in use only in mailroomapi and ppsgateway.  So, if you are willing to manually upgrade those repos, then this is ok.

artemchernyak-outreach (approved):

vk-outreach (approved):
LGTM given that the breaking change is limited to only a couple of repos.

A slightly different approach is to remove this option altogether and instead export two helpers:

```golang
func LogScheduler(s Scheduler, args ...log.Marshaler) Scheduler {
    ....  just export loggingScheduler
}

func LogRunner(r async.Runner, args ...log.Marshaler) async.Runner {
    ... just export a new runner which also logs on errors
}
```

I am ok with this specific change though.

pkg/async/pool/pool.go
artemchernyak-outreach: Sounds good to me. Thanks for looking into it!

artemchernyak-outreach (approved):

PR #187, opened 2022-07-18
feat: upgrade to latest bootstrap

captain-enjoyable:
Use latest bootstrap as part of bootstrap release process


bootstrap.lock
```diff
@@ -1,6 +1,6 @@
 # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
 # vim: set syntax=yaml:
-version: v9.37.0
-generated: 2022-05-18T19:22:35Z
+version: v10.2.0-rc.33
+generated: 2022-07-18T04:41:17Z
 versions:
-  devbase: v1.64.3
+  devbase: v2.1.0-rc.24
```
artemchernyak-outreach, line 6: Version 2 of devbase introduced the `--race` flag to all go tests. Looks like we didn't upgrade previously. 

This race condition either needs to be fixed or ignored with a follow up ticket to fix it.

To ignore the race condition: You can add the `RACE: disabled` env variable to the test step in `.circle-ci/config.yml`.

PR #188, opened 2022-07-18
feat(trace): deprecate functions

artemchernyak-outreach:

Deprecates functions that are no longer needed with the move to OpenTelemetry.

## Jira ID

[XX-XX]


ChrisPRobinson (approved):
LGTM

PR #189, opened 2022-07-19
feat: upgrade bootstrap to latest rc

artemchernyak-outreach:

Upgrades bootstrap to latest rc version.

## Jira ID

[DT-2146]

## Notes for your reviewers

[DT-2146]: https://outreach-io.atlassian.net/browse/DT-2146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/async/pool/pool.go
```diff
@@ -121,11 +121,12 @@ type Options struct {
 
 // Pool structure
 type Pool struct {
-	queue   chan unit
-	opts    *Options
-	wg      *sync.WaitGroup
-	cancel  func(error)
-	context context.Context
+	contextMu sync.Mutex
```
deregtd, line 124: it would be good to have a comment on exactly what the mutex is protecting, or if it's just protecting the entire struct.  contextMu to me implies it's protecting the context, but I see it used outside of just protecting the context.

artemchernyak-outreach: Let me leave a comment about that. It actually is just protecting the context because things like `orerr.CancelWithError` modify the context.

Let me also narrow down the locking a bit to make that more clear.

artemchernyak-outreach: Updated. That should clear things up a bit.

deregtd: ah i see now.  cancel is the method pulled from withcancel, so it's all attributes of the context.  as a tiny nit, i would organize the struct vars like that. but w/e.  thjx!

i.e.
```
contextMu
context
cancel

opts
queue
wg
```

deregtd (approved):
feedback in comments

PR #192, opened 2022-07-21
feat: [DWF-8223] Change Info to Debug

KirkJohnson:
[DWF-8223](https://outreach-io.atlassian.net/browse/DWF-8338) we have quite a few tracing logs in our services because we call multiple microservices and this ends up having a cost.  We'd like to convert these to Debug statements to reduce that cost.

If these aren't used for metrics we would be happy to have them removed.  We can't exclude the logs on our side without also excluding the errors which we do want to see since it seems DD exclusion logs only give us regex to do filtering by.

## Jira ID

[XX-XX]

### DIFF
--- a/pkg/trace/call.go
+++ b/pkg/trace/call.go
@@ -116,10 +116,10 @@ func EndCall(ctx context.Context) {
 			case statuscodes.CategoryServerError:
 				log.Error(ctx, info.Name, info, IDs(ctx), traceEventMarker{})
 			case statuscodes.CategoryOK: // just in case if someone will return non-nil error on success
-				log.Info(ctx, info.Name, info, IDs(ctx), traceEventMarker{})
+				log.Debug(ctx, info.Name, info, IDs(ctx), traceEventMarker{})
 			}
 		} else {
-			log.Info(ctx, info.Name, info, IDs(ctx), traceEventMarker{})
+			log.Debug(ctx, info.Name, info, IDs(ctx), traceEventMarker{})
 		}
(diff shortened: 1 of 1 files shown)


pkg/trace/call.go
```diff
@@ -116,10 +116,10 @@ func EndCall(ctx context.Context) {
 			case statuscodes.CategoryServerError:
 				log.Error(ctx, info.Name, info, IDs(ctx), traceEventMarker{})
 			case statuscodes.CategoryOK: // just in case if someone will return non-nil error on success
-				log.Info(ctx, info.Name, info, IDs(ctx), traceEventMarker{})
+				log.Debug(ctx, info.Name, info, IDs(ctx), traceEventMarker{})
 			}
 		} else {
-			log.Info(ctx, info.Name, info, IDs(ctx), traceEventMarker{})
+			log.Debug(ctx, info.Name, info, IDs(ctx), traceEventMarker{})
```
clevelittlefield-outreach, line 122: This would really mess with us and require a lot of change on our end.

clevelittlefield-outreach: Marking as request changes until we can discuss this further.

KirkJohnson: @clevelittlefield-outreach I think based on Slack thread we aren't going to do this.  This prompted discussion and hopefully a push towards design, but this doesn't seem like something Sync will own https://outreach-hq.slack.com/archives/C015GHLM5EK/p1658438757637789

PR #201, opened 2022-07-28
fix: only check ctx err once

artemchernyak-outreach:
## Jira ID

[DT-2254]

## Notes for your reviewers

[DT-2254]: https://outreach-io.atlassian.net/browse/DT-2254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/async/pool/pool.go
+++ b/pkg/async/pool/pool.go
@@ -196,7 +196,6 @@ func (p *Pool) run(ctx context.Context) {
 		case <-time.After(p.opts.ResizeEvery):
 			continue
 		case <-p.closed:
-			p.cancel(&orerr.ShutdownError{Err: context.Canceled})
 			return
 		case <-ctx.Done():
 			return
@@ -228,8 +227,9 @@ func (p *Pool) worker(ctx context.Context) {
 
 // Close blocks until all workers finshes current items and terminates
 func (p *Pool) Close() {
-	close(p.closed)
 	p.wg.Wait()
+	p.cancel(&orerr.ShutdownError{Err: context.Canceled})
+	close(p.closed)
 }
 
 // Schedule tries to schedule runner for processing in the pool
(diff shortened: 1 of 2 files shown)


kadukf-outreach (approved):

vk-outreach (approved):
LGTM.

pkg/async/pool/pool.go
```diff
@@ -166,7 +166,12 @@ func (p *Pool) run(ctx context.Context) {
 		prevSize, delta, size int
 		cancellations         = cancellations{}
 	)
-	for ctx.Err() == nil {
+
+	if ctx.Err() != nil {
+		return
+	}
+
+	for {
```
vk-outreach, line 174: I think it is still better to retain a `for ctx.Err() == nil { ... }` here because go does not guarantee that the switch statement will select the `<- ctx.Done()` clause immediately after a context is canceled -- it might pick up the retry etc.

artemchernyak-outreach:
Looks like this is still getting stuck in CI. Adding some notes on the updates that led to this state.

There were a couple race conditions found when the `--race` flag was added to `go test`:

- A race condition between the access to `ctx.Err` in the `run` loop and the update inside `Close`. To resolved this I moved the pool to listen to a `closed` channel instead of depending on `ctx.Err` to signal a shutdown.
- `TestPoolGrows` had a race condition when updating the pool size. To fix this I move `size` to be a channel that allowed communication between the pool resize goroutine and the main test goroutine.

Since the error is intermittent, one of the best ways I found to get the error locally (although I can’t seem to reproduce on the latest updates) was to run the tests inside `pkg/async/pool` 100 times. The below script is helpful in doing so:


```
for i in {1..100}
do
  result="$(go test -v --shuffle=on)"
  if echo $result | grep -q "FAIL:"
  then
    echo "$result"
  else
    echo "SUCCESS"
  fi
done
```

pkg/async/pool/pool.go
jaredallard: +1 to this

artemchernyak-outreach: Ah, thanks for the reminder on this.

edmangimelli-outreach (approved):

PR #202, opened 2022-08-01
feat: add lock to guard stdOut

LophyLiu:
We have been seeing random gobox/pkg/log [failures](https://app.circleci.com/pipelines/github/getoutreach/litmus/860/workflows/4fd62025-64b4-47b1-b848-5f854791409f/jobs/3523) in the CI. The root cause is `SetOutput()` and `Write()` are called concurrently trying to set/read the `stdOut` object at the same time. This change puts a lock around the `stdOut` object to guard it so that it is thread safe.

## Jira ID

[XX-XX]

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -42,9 +42,9 @@ var (
 	// wrap stdout and stderr in sync writers to ensure that writes exceeding
 	// PAGE_SIZE (4KB) are not interleaved.
 
-	stdOutLock *sync.RWMutex = new(sync.RWMutex)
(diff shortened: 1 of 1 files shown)


clevelittlefield-outreach (approved):

artemchernyak-outreach:
@LophyLiu There was another regression due to race conditions on main. That should be fixed now. Could you update this with main and then it's good to go.

artemchernyak-outreach (approved):

LophyLiu:
@artemchernyak-outreach The PR is updated with the latest change from main.

PR #204, opened 2022-08-03
fix: we don't want to wait for context to close the pool

pavelsmejkal:
## Jira ID

[XX-XX]

### DIFF
--- a/pkg/async/pool/pool.go
+++ b/pkg/async/pool/pool.go
@@ -227,9 +227,9 @@ func (p *Pool) worker(ctx context.Context) {
 
 // Close blocks until all workers finshes current items and terminates
 func (p *Pool) Close() {
-	p.wg.Wait()
 	p.cancel(&orerr.ShutdownError{Err: context.Canceled})
 	close(p.closed)
+	p.wg.Wait()
 }
 
 // Schedule tries to schedule runner for processing in the pool
--- a/pkg/async/pool/pool_test.go
+++ b/pkg/async/pool/pool_test.go
@@ -88,6 +88,10 @@ func (suite) TestGracefullyStops(t *testing.T) {
 	assert.Assert(t, InDelta(float64(s.NumGoroutineWithWorkers),
 		float64(runtime.NumGoroutine()), float64(size+1)), "Num of Goroutine is higher then expected")
 	s.Pool.Close()
+
+	// We don't want to wait for the context to close the pool.
+	assert.Assert(t, WithinDuration(time.Now(), s.StartedAt, 50*time.Millisecond))
+
 	time.Sleep(5 * time.Millisecond)
 	// After close all workers goroutines are dead
 	assert.Assert(t, InDelta(float64(s.NumGoroutineOnStart),


kadukf-outreach (approved):

jaredallard (approved):

pkg/async/pool/pool.go
```diff
@@ -227,9 +227,9 @@ func (p *Pool) worker(ctx context.Context) {
 
 // Close blocks until all workers finshes current items and terminates
 func (p *Pool) Close() {
-	p.wg.Wait()
 	p.cancel(&orerr.ShutdownError{Err: context.Canceled})
```
artemchernyak-outreach, line 230: @pavelsmejkal canceling the pool before the wait group completes was the cause of the original race condition in this library. If this is the order that we need to do things in, we should probably use a mutex to ensure that `ctx.Err` doesn't cause a race condition.

artemchernyak-outreach (changes requested):

pkg/async/pool/pool.go
pavelsmejkal: @artemchernyak-outreach, true i was just leaving the office and didn't wait for all tests. But then it is not a problem of the pool but of the CancelWithError. Submitted the fix

pkg/async/pool/pool_test.go
```diff
@@ -119,21 +114,47 @@ func (suite) TestPoolGrows(t *testing.T) {
 		}),
 		ResizeEvery: 1 * time.Millisecond,
 	}
-	ng = runtime.NumGoroutine()
-
 	runPool(context.Background(), s)
 
 	defer s.Cancel()
```
kadukf-outreach, line 118: @pavelsmejkal do I need to cancel here?

Shouldn't it be enough to close the pool -> with current fix it should work and prove that closing the pool really stops the pool

pkg/async/pool/pool_test.go
```diff
@@ -79,30 +76,27 @@ func (suite) TestWeCantEnqueueWhenStopped(t *testing.T) {
...
+
+	assert.Assert(t, waitForWorkers(t, 0), "workers detected")
 }
 
 // TestPoolGrows checks number of running goroutines can't be execute using shuffler that run tests in parallel
-func (suite) TestPoolGrows(t *testing.T) {
+func TestPoolGrows(t *testing.T) {
 	var size = make(chan int, 1)
-	ng := 0
+	//ng := 0
```
kadukf-outreach, line 94: nit: remove the comment

pkg/async/pool/pool_test.go
pavelsmejkal: i have removed it

pkg/async/pool/pool_test.go
pavelsmejkal: sure

pavelsmejkal:
Sorry for all the message, tests were passing on my computer but not on CI.

artemchernyak-outreach (approved):

artemchernyak-outreach:
I really like the new test updates. Thanks for adding those in!

PR #209, opened 2022-08-16
feat: support additional endpoint trace config

artemchernyak-outreach:

Allows configuring additional endpoints and keys so that we can support multiple trace backends at once.

## Jira ID

[DT-2637]

## Notes for your reviewers

[DT-2637]: https://outreach-io.atlassian.net/browse/DT-2637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/trace/config.go
+++ b/pkg/trace/config.go
@@ -21,15 +21,24 @@ func (g *GlobalTags) MarshalLog(addField func(key string, v interface{})) {
 }
 
 type Otel struct {
-	Enabled            bool       `yaml:"Enabled"`
-	Endpoint           string     `yaml:"Endpoint"`
-	AdditionalEndpoint string     `yaml:"AdditionalEndpoint"`
-	Dataset            string     `yaml:"Dataset"`
-	SamplePercent      float64    `yaml:"SamplePercent"`
-	Debug              bool       `yaml:"Debug"`
-	Stdout             bool       `yaml:"Stdout"`
-	APIKey             cfg.Secret `yaml:"APIKey"`
-	AdditionalAPIKey   cfg.Secret `yaml:"AdditionalAPIKey"`
+	// Enabled determines whether to turn on tracing
+	Enabled bool `yaml:"Enabled"`
+	// Endpoint for the tracing backend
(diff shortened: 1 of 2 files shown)


pkg/trace/config.go
```diff
@@ -21,13 +21,15 @@ func (g *GlobalTags) MarshalLog(addField func(key string, v interface{})) {
...
 
 type Otel struct {
-	Enabled       bool       `yaml:"Enabled"`
-	Endpoint      string     `yaml:"Endpoint"`
-	Dataset       string     `yaml:"Dataset"`
-	SamplePercent float64    `yaml:"SamplePercent"`
-	Debug         bool       `yaml:"Debug"`
-	Stdout        bool       `yaml:"Stdout"`
-	APIKey        cfg.Secret `yaml:"APIKey"`
+	Enabled            bool       `yaml:"Enabled"`
```
jaredallard, line 24: nit: Can we get a comment on each of these, lintroller plat will want it in the future

pkg/trace/otel.go
```diff
@@ -104,16 +104,50 @@ func (t *otelTracer) initTracer(ctx context.Context, serviceName string) error {
...
+			sdktrace.WithResource(r),
+			// accepts sample rates as number of requests seen per request sampled
+			sdktrace.WithSampler(forceSample(uint(100/t.Otel.SamplePercent))),
+			sdktrace.WithSpanProcessor(Annotator{
+				globalTags: t.GlobalTags,
+				sampleRate: int64(100 / t.Otel.SamplePercent),
+			}),
+		)
+	} else {
+		tp = sdktrace.NewTracerProvider(
```
jaredallard, line 140: Could we combine this and L128 by making the options a slice that we add to instead? I don't really like duplicating the code here :'(

artemchernyak-outreach: That's totally fair. Will do.

jaredallard (approved):

PR #212, opened 2022-08-20
feat(updater): rewrite updater to better support channels

jaredallard:

This PR rewrites the updater to better support multiple release channels, more archive types, and be more stable / tested.

## Jira ID

[XX-XX]


pkg/updater/resolver/resolver_test.go
```diff
@@ -0,0 +1,187 @@
...
+			c:    Criteria{Channel: "rc"},
+			versions: map[string][]Version{
+				StableChannel: {
+					newTestingVersion("v1.0.0", false),
+				},
+				"rc": {
+					newTestingVersion("v0.9.0", true),
+				},
+			},
+			want:    newTestingVersion("v0.9.0", true),
```
jaredallard, line 95: Note to self: This is wrong, want should be v1, fix logic & test

jaredallard:
I'm going to iterate on the config package _again_ because the new API I setup is better, but more confusing to use for really no reason. However, I feel like this is ready for basic review so, take a look!

I've been testing this on bootstrap with great success, and I'm pretty happy with the test coverage. Note that the current test coverage of the `updater` package is only 40%, this is mostly because I don't have a good way to cover the urfave/cli parts of this library, however using a `replace` in a CLI is an easy way to test it, so I'm content with local testing of that for now.

jaredallard:
This is 100% a breaking change. I can add back the old `updater` path and just deprecated all the methods, if we want. That'd prevent breaking the API surface. LMK @artemchernyak-outreach / @george-e-shaw-iv

pkg/cli/updater/README.md
```diff
@@ -0,0 +1,26 @@
...
+ * Support multiple different CLI frameworks entry points, e.g. `urfave/cli` and cobra.
+
+
+### Version Logic
+
+ * If the user's config says to be on a certain channel, consider that channel (note: bullet 3 applies)
+ * If on a channel (non-release version) we should automatically consider versions from that channel for updating.
+ * Release versions should be ignored if the user is on the `rc` or `unstable`, or X other channel UNLESS the release
+   version is greater than the current version. 
+    * **Note**: This means if you want an LTS branch, you'll have to make "stable" that LTS branch. Always add more lower release channels instead of n above stable.
```
artemchernyak-outreach, line 25: Consider including a few examples of how to set this up and migrate.

It was very hard to try to figure out how to migrate things.

artemchernyak-outreach (approved):
Code and tests look great! Just one suggestion to make it easier to divide and conquer the migration to the new updater.

pkg/cli/updater/README.md
jaredallard: Migration should largely have no impact, when we moved the bootstrap cli code into `gobox` that moved the only call to the updater.

However, some applications (bootstrap, for example) have been using the updater for awhile now to download other applications by using the undocumented/accidentally exported github client (sorry). Those will require some sort of move, most likely, to use the new sub-packages. I don't think there will be that many, so I'll clean those up.

Thinking about it more, I think there really isn't a breaking change here...

jaredallard:
Not signed because I rebased via the UI

PR #217, opened 2022-08-28
feat(cli): Better shell completion support.

jkinkead:

Eject into shell completion mode immediately, without completing logging and tracing setup. This dramatically increases the responsiveness of shell completion, and also avoids injecting updater messages into the completion suggestions.

Add a special flag for fish completion, since there is none built-in for urfave.

## Notes for your reviewers

See [the urfave manual](https://cli.urfave.org/v2/#bash-completion) for details on bash / zsh / powershell completion. The TL;DR is urfave supplies [helper scripts](https://github.com/urfave/cli/tree/8b23e7b1e99f37934e029ef6c8d31efc7e744ff1/autocomplete) that you can source to make completion work.

Fish support [exists](https://github.com/urfave/cli/blob/8b23e7b1e99f37934e029ef6c8d31efc7e744ff1/fish.go), but is undocumented in the manual and kind of wonky, since you must invoke it separately. Additionally, the implementation creates a completion script that would be sourced once at shell startup, since that allows for faster completion (you don't need to invoke the urfave binary again).

### DIFF
--- a/pkg/cli/bootstrap.go
+++ b/pkg/cli/bootstrap.go
@@ -162,12 +162,67 @@ func setupExitHandler(ctx context.Context) (exitCode *int, exit func()) {
 	return
 }
 
+// Returns true if the last argument to the program is
+// --generate-bash-completion, the special hidden flag urfave creates.
+// See: https://cli.urfave.org/v2/#bash-completion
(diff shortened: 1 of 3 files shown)


artemchernyak-outreach (approved):
Great idea! LGTM

PR #218, opened 2022-08-29
feat: CSV reader

dundee:

CSV reader forker from std `encoding/csv` with removed `\r\n` handling, so it preserves the original content without modifications.

https://github.com/golang/go/issues/22746

## Jira ID

[RMS-983]

## Notes for your reviewers

[RMS-983]: https://outreach-io.atlassian.net/browse/RMS-983?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deregtd (changes requested):
talked offline about this -- should be a forked package properly, or switch to another package/etc. -- we want gobox to be outreach code

PR #219, opened 2022-09-08
feat(updater/resolver): support constraints (DT-2120)

jaredallard:

This PR adds [constraint support](https://pkg.go.dev/github.com/Masterminds/semver/v3#hdr-Checking_Version_Constraints_and_Comparing_Versions) to the version resolver used by the updater. This allows other consumers to add semver constraints to change the versions selected by the resolver.

The general behaviour of this package is well documented in the tests written, but also in the package comment for `Resolve`, copied here for posterity:

```go
// Resolve returns the latest version that satisfies the criteria
//
// Note: token is _optional_, if you do not wish to authenticate with your VCS
// provider you can pass an empty string.
//
// Criteria:
//
// Constraints are supported as per the underlying semver library here:
// https://pkg.go.dev/github.com/Masterminds/semver/v3#hdr-Checking_Version_Constraints_and_Comparing_Versions
//
// If a channel is provided, without a constraint, the latest version of that channel will be returned if it's
// greater than the `stable` channel (if present). As a special case, if a channel's latest version is mutable,
// that version will always be returned over any constraint.
//
// Constraints, by default, do not include a pre-releases or allow selecting specific pre-releases tracks in the
// semver library used. In order to support this we parse the constraints to determine the allowed channels. By
// default `stable` is always considered, with whatever pre-release (e.g. `>=1.0.0-alpha` -> `alpha`) is specified
// in the pre-release constraint being added to the allowed channels. If a channel is specified, it is also added
// to the allowed channels.
func Resolve(ctx context.Context, token cfg.SecretData, c *Criteria) (*Version, error) {
```

## Jira ID

[DT-2120]

## Notes for your reviewers

[DT-2120]: https://outreach-io.atlassian.net/browse/DT-2120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/cli/updater/resolver/resolver.go
+++ b/pkg/cli/updater/resolver/resolver.go
@@ -203,7 +203,7 @@ func getVersions(ctx context.Context, token cfg.SecretData, url string) (map[str
 // semver library used. In order to support this we parse the constraints to determine the allowed channels. By
 // default `stable` is always considered, with whatever pre-release (e.g. `>=1.0.0-alpha` -> `alpha`) is specified
 // in the pre-release constraint being added to the allowed channels. If a channel is specified, it is also added
-// to the allowed channels.
+// to the allowed channels. Due to this '&&' and '||' are not supported in constraints.
 func Resolve(ctx context.Context, token cfg.SecretData, c *Criteria) (*Version, error) {
 	versions, err := GetVersions(ctx, token, c.URL)
 	if err != nil {
--- a/pkg/cli/updater/resolver/resolver.go
+++ b/pkg/cli/updater/resolver/resolver.go
@@ -12,8 +12,9 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"strings"
 
-	"github.com/blang/semver/v4"
+	"github.com/Masterminds/semver/v3"
 	"github.com/getoutreach/gobox/pkg/cfg"
 	"github.com/go-git/go-git/v5"
 	gitconfig "github.com/go-git/go-git/v5/config"
@@ -28,6 +29,15 @@ import (
 // https://regex101.com/r/yWXH5i/1
 var mutableTagRegex = regexp.MustCompile(`^[a-z-]+$`)
 
+// oprStripRegex is the regex to use to strip operators from a constraint
+var oprStripRegex = regexp.MustCompile(`^\D+(\d)`)
+
+// This block contains error types for the resolver package.
+var (
+	// ErrNoVersions is returned when no versions are found
+	ErrNoVersions = errors.New("no version found matching criteria")
+)
+
 // StableChannel is the default channel used when a version
 // doesn't contain a channel.
 const StableChannel = "stable"
@@ -40,6 +50,10 @@ type Criteria struct {
 	// Channel is the channel to use for determining the latest version.
 	// Leave empty to use the StableChannel.
 	Channel string
+
+	// Constraints are the semver constraint(s) to use for determining the latest version.
+	// See: https://pkg.go.dev/github.com/Masterminds/semver/v3#hdr-Checking_Version_Constraints_and_Comparing_Versions
+	Constraints []string
 }
 
 // Version is a resolved version from a git repository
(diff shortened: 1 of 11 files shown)


pkg/cli/updater/resolver/resolver.go
```diff
@@ -73,6 +87,44 @@ func (v *Version) String() string {
 	return v.Tag
 }
 
+// NewVersion creates a new version from a tag and commit.
+func NewVersion(tag, hash string) (*Version, error) {
```
jaredallard, line 91: This isn't really related to this PR, but it allowed me to cleanup some code duplication.

pkg/cli/updater/resolver/resolver.go
```diff
@@ -181,33 +219,155 @@ func Resolve(ctx context.Context, token cfg.SecretData, c *Criteria) (*Version,
...
+		// strip space from the constraint
+		constraintStr = strings.TrimSpace(constraintStr)
+		if constraintStr == "*" {
+			// for some reason * doesn't work with the prereleases hack we use below,
+			// so convert * into >= 0.0.0, it's equivalent.
+			constraintStr = ">=0.0.0"
+		}
+
+		// extract allowed channels from the constraint if there is one
+		exampleVer := oprStripRegex.ReplaceAllString(constraintStr, "$1")
```
jkinkead, line 261: Why have a capture at all if you're just replacing a single character? Just replacing `^\D+` with `""` should suffice.

jaredallard: Nuance of go regex, we want to ensure that there is a number after the non-digit, but in order to do that we have to match... which means it'll be replaced. So, we capture so we can add it back :/

jkinkead: Oh, you want to ensure it's non-empty. Use a non-capturing group:
```
^\D(?:\d)
```

jkinkead: Wait, nm - that'll still get matched. You want a zero-width lookahead, but it looks like go doesn't support that?

jaredallard: Yep :(

pkg/cli/updater/resolver/resolver.go
```diff
@@ -181,33 +219,155 @@ func Resolve(ctx context.Context, token cfg.SecretData, c *Criteria) (*Version,
...
+				}
 			}
 		}
+
+		// If we're allowing channels other than the stable channel then we need to
+		// mutate the constraint to allow pre-releases. The constraint matching doesn't
+		// allow you to specify which pre-releases to allow, so we just allow all here
+		// and filter it down later.
+		if c.Channel != StableChannel && !strings.Contains(constraintStr, "-") {
+			constraintStr += "-" + "prereleases"
```
george-e-shaw-iv, line 278: ...? two literals
```suggestion
			constraintStr += "-prereleases"
```

george-e-shaw-iv (approved):
from a reviewers POV it looks fine, but this is definitely hard to grok without being the author. the test coverage on your additive lines make me feel better so i'll leave an approval

pkg/cli/updater/resolver/resolver_test.go
```diff
@@ -58,42 +38,167 @@ func TestResolve(t *testing.T) {
...
+				},
+				"rc": {
+					newTestingVersion("v0.9.1-rc.1"),
+				},
+				"beta": {
+					newTestingVersion("v0.9.2-beta.1"),
+				},
+			},
+			// should return beta because channel asked for it
+			want: newTestingVersion("v0.9.2-beta.1"),
```
artemchernyak-outreach, line 201: This looks great! Just to cover all test cases, we should add a test where we compare two rc versions (e.g. `v0.9.1-rc.1` and `v0.9.1-rc.2`).

artemchernyak-outreach (approved):

PR #221, opened 2022-09-09
fix(trace): export SpanID

nissim-natanov-outreach:
This PR exposes SpanID as well as Trace and Span ID from the Link object. It also exports the former linkOption as a Link struct to allow callers access the trace/span IDs directly from the link.

Clerk needs to link a trace to process batch of clerk events to their producer's traces. If producer sent them in a batch, there is a good chance that there will be large overlaps between producer and consumer batches with span IDs, resulting in multiple links with same trace/span ID.

I checked otel's code - it does not deduplicate links. If the # of links is over limit (default 128), it will simply evict others.
https://github.com/open-telemetry/opentelemetry-go/blob/55b49c407e07dc13bdba245c2c13935a37bad971/sdk/trace/tracer.go#L144. SpanID is needed to deduplicate the links.

## Jira ID

[CDT-102]

### DIFF
--- a/pkg/trace/call.go
+++ b/pkg/trace/call.go
@@ -148,7 +148,7 @@ type traceInfo struct {
 func (c traceInfo) MarshalLog(addField func(field string, value interface{})) {
 	addField("honeycomb.trace_id", ID(c))
 	addField("honeycomb.parent_id", parentID(c))
-	addField("honeycomb.span_id", spanID(c))
+	addField("honeycomb.span_id", SpanID(c))
 }
 
 type traceEventMarker struct{}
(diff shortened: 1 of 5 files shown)


artemchernyak-outreach:
The deduping is a great strategy. In addition, I don't think we have any concrete reason to keep a link limit. Looks like you can make it unlimited by setting it to a negative number (as per https://github.com/open-telemetry/opentelemetry-go/blob/fdfc821bac8357ec1eac271c50454606c318d8de/sdk/trace/span_limits.go#L78).

I think that would be a good addition here.

nissim-natanov-outreach:
> The deduping is a great strategy. In addition, I don't think we have any concrete reason to keep a link limit. Looks like you can make it unlimited by setting it to a negative number (as per https://github.com/open-telemetry/opentelemetry-go/blob/fdfc821bac8357ec1eac271c50454606c318d8de/sdk/trace/span_limits.go#L78).
> 
> I think that would be a good addition here.

Thanks Artem, our default batch is 10,  I think it will take time until people adapt batching and master it, for us to have enough real traces in honeycomb to play around. Once there, we can decide if we need to increase the limits or simply set the link count to infinite - I am not sure yet what the consequences to cost and (mostly) usability of the links are. Prob worth doing so in separate PRs

artemchernyak-outreach (approved):

jaredallard (approved):

jkinkead (approved):

PR #222, opened 2022-09-11
chore: Unit test fixes.

jkinkead:

Fix a test whose slow task would sometimes complete too early, breaking a timeout test.

Fix a test that would sometimes cancel too early, causing too few tasks to finish. Also load shared memory atomically, just for completeness.

Clean up some grammar in godoc.

## Notes for your reviewers

These tests are super-flaky, and hopefully this fixes them.

### DIFF
--- a/pkg/async/pool/timeout_test.go
+++ b/pkg/async/pool/timeout_test.go
@@ -14,32 +14,38 @@ func TestTimeout(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
-	p := pool.New(ctx,
-		pool.ConstantSize(1),
-	)
+	p := pool.New(ctx, pool.ConstantSize(1))
 
 	startedAt := time.Now()
 
-	scheduler := pool.WithTimeout(5*time.Millisecond, p)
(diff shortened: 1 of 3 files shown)


pkg/async/pool/timeout_test.go
```diff
@@ -14,32 +14,38 @@ func TestTimeout(t *testing.T) {
...
+	defer close(startFirst)
 
 	scheduler.Schedule(ctx, async.Func(func(ctx context.Context) error {
 		if ctx.Err() != nil {
 			return ctx.Err()
 		}
-		time.Sleep(10 * time.Millisecond)
+		<-startFirst
+		// Ensure the next task times out.
+		time.Sleep(2 * timeout)
```
artemchernyak-outreach, line 37: Great idea on ensuring this timeout!

artemchernyak-outreach (approved):

pkg/async/pool/timeout_test.go
jkinkead: You complementing the variable? I'm actually generally terribly lazy when it comes to tests, but this helped the readability here. :)

jaredallard (approved):

PR #230, opened 2022-09-26
feat: handle invalid semver, support go run/install/gobin

jaredallard:

This PR makes the updater disable itself when an invalid version is provided. I also updated `app` to source version information, if not linked in, from the go module. This lets `go run` and (theoretically, I can't test w/o a release) work on `go install` (and `go run <module>@<revision>`). Unclear if this will support gobin, but it'll at least make `--skip-update` a valid argument.

Example on a locally compiled version of `smartstore` w/ `replace` for this version:

```bash
$ go run ./cmd/smartstore updater status
Version: (devel)
Channel:  ()
Updater Status: disabled (failed to parse current version as semver: Invalid Semantic Version)
Last Update Check: Mon, 26 Sep 2022 18:11:52 UTC
```

## Jira ID

[DT-2907]

## Notes for your reviewers

[DT-2907]: https://outreach-io.atlassian.net/browse/DT-2907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/cli/updater/updater.go
+++ b/pkg/cli/updater/updater.go
@@ -133,11 +133,6 @@ func (u *updater) defaultOptions() error {
 		u.version = app.Info().Version
 	}
 
-	curVersion, err := semver.NewVersion(u.version)
-	if err != nil {
-		return errors.Wrapf(err, "failed to parse current version %q as semver", u.version)
-	}
-
 	if u.executablePath == "" {
 		var err error
 		u.executablePath, err = exec.ResolveExecuable(os.Args[0])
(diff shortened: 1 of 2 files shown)


pkg/cli/updater/updater.go
```diff
@@ -188,16 +183,24 @@ func (u *updater) defaultOptions() error {
...
 	}
 
-	// determine channel from version string if not set
-	if u.channel == "" {
-		u.channel = curChannel
-		u.channelReason = fmt.Sprintf("using %s version", curChannel)
+	if curVersion != nil {
+		curChannel, curLocalBuild := u.getVersionInfo(curVersion)
+		if curLocalBuild {
+			u.disabled = true
```
artemchernyak-outreach, line 195: nit: it would be nice to add a test for this since we already have the test suite. Other than that, LGTM.

artemchernyak-outreach (approved):

pkg/cli/updater/updater.go
jaredallard: I would, but I don't know a good way to do that in tests currently. I can build a fake out for `debug.ReadBuildInfo`, but that still doesn't help me all that much because I don't know it's behaviour all that well right now. Was hoping to refine as we use it more (it really will only impact go run/install, which is already broke atm)

artemchernyak-outreach: Definitely makes sense.

AndrewWinterman (approved):
lgtm

PR #231, opened 2022-09-27
feat: add generic priority queue

LophyLiu:
This PR adds a generic priority queue implementation so that people can take it and use without re-implementing the priority queue. Priority queue implements the `container/heap` and enriches it with additional features, such as Peek, Update, Remove, Contains, List, Clear and etc.


pkg/queue/error.go
```diff
@@ -0,0 +1,27 @@
...
+// NewMaxCapacityError
+func NewMaxCapacityError(capacity int) *MaxCapacityError {
+	return &MaxCapacityError{
+		capacity: capacity,
+	}
+}
+
+var _ error = new(MaxCapacityError)
+
+// MaxCapacityError
```
jaredallard, line 19: Please provide more descriptive comments than just the including the function name to 'game' the linter, these should be indicative of why/what this is. Consider it to be justifying its existence. 

This applies to all comments in this PR.

pkg/queue/priority_queue.go
```diff
@@ -0,0 +1,210 @@
...
+func NewPriorityQueue(opts ...PriorityQueueOption) PriorityQueue {
+	q := &priorityQueue{
+		capacity: defPriorityQueueCapacity,
+		queue:    new(priorityQueueInternal),
+	}
+	for _, opt := range opts {
+		opt(q)
+	}
+	if q.capacity <= 0 {
+		panic(errors.Errorf("invalid capacity: %v", q.capacity))
```
jaredallard, line 66: Generally never panic in a library, return an error instead

pkg/queue/priority_queue_test.go
```diff
@@ -0,0 +1,142 @@
+package queue
```
jaredallard, line 1: Can we add some examples here as well so that there's examples in the generated function docs?

pkg/queue/error.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+// NewMaxCapacityError
+func NewMaxCapacityError(capacity int) *MaxCapacityError {
+	return &MaxCapacityError{
+		capacity: capacity,
+	}
+}
+
+var _ error = new(MaxCapacityError)
```
jaredallard, line 17: Missing comment here

```suggestion
// _ ensures that MaxCapactityError implements the error interface
var _ error = new(MaxCapacityError)
```

pkg/queue/error.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
+//
+// Description: error
```
jaredallard, line 3: Same thing for the description here, it should go into detail about why this file exists.

jaredallard (changes requested):
Left some review comments, I'd like to see a lot more details on this PR in both the comments and the description justifying why we should be adding this to our open source core library and what the benefits it brings are.

I'd like to see the following:

 * Comments addressed, I consider the ones left here to be 'required', though I'm happy to chat more about them 
 *  A detailed description of this in both the package comment and the PR description, why it exists, why this brings benefits over `container/heap`'s example and being an interface
 * More comments on the logic being implemented on top of `container/heap`

pkg/queue/priority_queue.go
LophyLiu: ok

pkg/queue/priority_queue_test.go
```diff
@@ -0,0 +1,142 @@
...
+	queue := NewPriorityQueue()
+	ast.Equal(0, queue.Len())
+	ast.Nil(queue.Peek())
+	ast.Nil(queue.Pop())
+
+	n := 1000
+	repeat := 3
+	for i := 0; i < repeat; i++ {
+		// push numbers in random order
+		pushRandN2PiroirtyQueue(ast, queue, n)
```
clevelittlefield-outreach, line 21: Typo

pkg/queue/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
+//
+// Description: queue data struct.
+
+// Package queue provides queue data struct.
+package queue
```
clevelittlefield-outreach, line 6: This feels like the wrong package name

pkg/queue/priority_queue.go
```diff
@@ -0,0 +1,210 @@
...
+func (q *priorityQueue) Peek() PriorityQueueIndexItem {
+	if q.Len() <= 0 {
+		return nil
+	}
+	return q.queue.Peek()
+}
+
+// Update
+func (q *priorityQueue) Update(item PriorityQueueIndexItem) {
+	heap.Fix(q.queue, item.getIndex())
```
clevelittlefield-outreach, line 110: Shouldnt this be updating on priority, not index?

pkg/queue/priority_queue_test.go
```diff
@@ -0,0 +1,142 @@
+package queue
+
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
clevelittlefield-outreach, line 7: testify is not really used in this library

LophyLiu: `go.sum` is not changed, meaning gobox has an indirect dependency on `testify` already. I see [gotest.tools/v3/assert](https://github.com/gotestyourself/gotest.tools) is being used, but given [testify](https://github.com/stretchr/testify) is more popular and richer than `gotest.tools/v3/assert`, what is the problem of using `testify`?

pkg/queue/priority_queue_test.go
LophyLiu: thx.

pkg/queue/doc.go
clevelittlefield-outreach: If the reviewers here are ok with it, but when I think queue I would expect a standard queue implementation first.  Maybe this should be data_structures or whatever is go idiomatic?

pkg/queue/priority_queue_test.go
clevelittlefield-outreach: testify is indirectly referenced but if you look the entire rest of the unit tests in this file contain no testify-based assertions.  You have to adapt to the repo, and cannot just introduce a new testing paradigm or strengthen ties to an indirect dependency.

LophyLiu: I don't think sticking with `gotest` is a good idea. Look at some of the existing gobox test codes. Some are creating assertion helpers (DeepEqual, InDelta,  WithinDuration) which are not reusable across packages, some are creating suite and before/after test. All of these are available in `testify` already. I think it is better to move to a better testing library rather than writing repeated test helpers with `gotest` that are not reusable across packages.

* TestSuite
	* [gotest](https://github.com/getoutreach/gobox/blob/v1.54.0/pkg/log/fatal_test.go#L17)
	* [testify suite](https://pkg.go.dev/github.com/stretchr/testify@v1.8.0/suite)
* Before/AfterTest
	* [gotest](https://github.com/getoutreach/gobox/blob/v1.54.0/pkg/cli/github/auth_test.go#L22-L23)
	* [testify](https://pkg.go.dev/github.com/stretchr/testify@v1.8.0/suite#BeforeTest)
* DeepEqual
	* [gotest](https://github.com/getoutreach/gobox/blob/v1.54.0/pkg/log/caller_test.go#L31)
	* [testify EquvalValues](https://pkg.go.dev/github.com/stretchr/testify@v1.8.0/assert#EqualValues)
* Delta
	* [gotest](https://github.com/getoutreach/gobox/blob/v1.54.0/pkg/async/pool/utils_test.go#L10-L34)
	* [testify](https://pkg.go.dev/github.com/stretchr/testify@v1.8.0/assert#Assertions.InDelta)

pkg/queue/doc.go
LophyLiu: There could be other queue implementation with any/combination of the following characteristics, they all fit in the queue package.
* Blocking: block on push/pop
* FIFO: first in first out
* TTL: item will TTL
* Stream: in memory kafa
* Delay: item in queue is available with a delay

clevelittlefield-outreach: I understand different queue types, I am just saying right now we do not have those and it is weird to come to a queue package that has one queue that is a more esoteric queue type.  All of those are likely backed by a real external queue system and the implementation here would just be a wrapper and likely not right for gobox.

pkg/queue/priority_queue_test.go
LophyLiu: Moving this conversation to [here](https://outreach-hq.slack.com/archives/C01CAR6GVPD/p1664315615985359).

pkg/queue/doc.go
LophyLiu: Any naming recommendation? Or moving this package to `queue/priorityqueue`?

LophyLiu: Will use `queue` for now, unless another package name is proposed.

LophyLiu:
@jaredallard @clevelittlefield-outreach , pr is updated. please review at your connivence.

pkg/queue/priority_queue_test.go
```diff
@@ -0,0 +1,344 @@
...
+	assert.Assert(t, err == nil)
+	assert.Assert(t, queue.Len() == 0)
+	assert.Assert(t, queue.Peek() == nil)
+	assert.Assert(t, queue.Pop() == nil)
+
+	n := 1000
+	repeat := 3
+	for i := 0; i < repeat; i++ {
+		// push numbers in random order
+		pushRandNumbersToPiroirtyQueue(t, queue, n)
```
clevelittlefield-outreach, line 23: still typo

LophyLiu: no sure how I miss it, but good catch.

pkg/queue/priority_queue.go
```diff
@@ -0,0 +1,366 @@
...
+import (
+	"container/heap"
+	"math"
+	"sort"
+	"sync"
+
+	"github.com/pkg/errors"
+)
+
+// const
```
jaredallard, line 59: Comment should justify why there's a group of consts here, since you only have one I suggest just doing `const Def...`

jaredallard (approved):
LGTM after last piece of feedback

pkg/queue/priority_queue.go
LophyLiu: done.

george-e-shaw-iv (approved):

PR #234, opened 2022-10-12
fix(updater): fix updating binaries

samnguyen-outreach:
Where the binary is not the same name as the repo

- Uses repoName instead of executableName
- Executable is not always named the same as the repo, but CI uploads assets that are the name of the repo.

## Jira ID

[XX-XX]


jaredallard (approved):
Thank you!!

AndrewWinterman (approved):

PR #236, opened 2022-10-13
feat(logfile): add automatic logfile recorder/player for CLIs

jaredallard:

This PR adds an automatic cli logfile generator as well as player for said log files

I also removed the current tracing and logging code in the current CLI helpers, as well as better organized it. The logfile work is contained to `pkg/cli/logfile`

## Jira ID

[DT-0]


pkg/cli/cli.go
```diff
@@ -0,0 +1,85 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
+
+// Description: See package comment
```
jkinkead, line 3: This is for the file contents, isn't it? This package has multiple files in it - shouldn't this describe the contents of this file?

pkg/cli/shutdown.go
```diff
@@ -0,0 +1,50 @@
...
+
+		// Go sets panic exit codes to 2
+		(*exitCode) = 2
+	}
+}
+
+// setupExitHandler sets up an exit handler that will call os.Exit() with
+// the set exit code, ensuring that all log/trace data is flushed.
+func setupExitHandler() (exitCode *int, exit func()) {
+	exitCode = intPtr(0)
```
jkinkead, line 43: Can't you just use a local and get a pointer to that?
```suggestion
	exitCodeInt := 0
	exitCode := &exitCodeInt
```

jkinkead (commented):
Two short comments, may review fully later ...

pkg/cli/cli.go
jaredallard: It is, but the package comment describes what's in this file. I can duplicate it if you'd prefer

pkg/cli/shutdown.go
jaredallard: Yeah, this is just copied code. Will adjust since i need to remove that function as well

artemchernyak-outreach (approved):

pkg/cli/cli.go
jkinkead: Shouldn't you fix the package comment, then?

jaredallard: Yeah, I guess it was slightly messed up. Should be better now. LMK

jaredallard:
Closing until this can be taken over again :(

PR #238, opened 2022-10-18
feat: logfile traces

artemchernyak-outreach:
## Jira ID

[XX-XX]

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -8,8 +8,13 @@ package cli
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
+	osexec "os/exec"
+	"os/user"
+	"runtime"
+	"strings"
 
 	"github.com/getoutreach/gobox/pkg/app"
 	"github.com/getoutreach/gobox/pkg/cli/logfile"
@@ -52,10 +57,15 @@ func HookInUrfaveCLI(ctx context.Context, cancel context.CancelFunc, a *cli.App,
 
 	// Support loading compiled in keys from the binary through the
 	// config framework
-	overrideConfigLoaders(honeycombAPIKey, dataset, false)
+	overrideConfigLoaders()
 
 	// Cancel the context on ^C and other signals
 	urfaveRegisterShutdownHandler(cancel)
+
+	ctx = setupTracer(ctx, a.Name)
+
+	props := commonProps()
+	trace.AddInfo(ctx, props)
 
 	exitCode, exit := setupExitHandler()
 	defer exit()
@@ -77,4 +87,44 @@ func HookInUrfaveCLI(ctx context.Context, cancel context.CancelFunc, a *cli.App,
 
 		return
 	}
+}
+
+// commonProps sets up common properties for traces in cli apps
+func commonProps() log.Marshaler {
+	commonProps := log.F{
+		"os.name": runtime.GOOS,
+		"os.arch": runtime.GOARCH,
+	}
+	if b, err := osexec.Command("git", "config", "user.email").Output(); err == nil {
+		email := strings.TrimSuffix(string(b), "\n")
+
+		// TODO: Turn the check into an config option
+		// In case of @outreach.io email, we want to add PII for easier debugging with devs
+		if strings.HasSuffix(email, "@outreach.io") {
+			commonProps["dev.email"] = email
+
+			if u, err := user.Current(); err == nil {
+				commonProps["os.user"] = u.Username
+			}
+
+			if hostname, err := os.Hostname(); err == nil {
+				commonProps["os.hostname"] = hostname
+			}
+			path, err := os.Getwd()
+			if err == nil {
+				commonProps["os.workDir"] = path
+			}
+		}
+	}
+
+	return commonProps
+}
+
+// setupTracer initializes tracing for clis
+func setupTracer(ctx context.Context, name string) context.Context {
+	if err := trace.InitTracer(ctx, name); err != nil {
+		fmt.Println(err)
+		return ctx
+	}
+	return trace.StartSpan(ctx, name)
 }
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -8,13 +8,8 @@ package cli
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"os"
-	osexec "os/exec"
-	"os/user"
-	"runtime"
-	"strings"
 
 	"github.com/getoutreach/gobox/pkg/app"
 	"github.com/getoutreach/gobox/pkg/cli/logfile"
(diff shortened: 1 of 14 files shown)


pkg/cli/cli.go
```diff
@@ -8,8 +8,13 @@ package cli
```
jaredallard, line 8: This should probably mostly be in `trace.go` or something like that, to keep the file small

pkg/cli/config.go
```diff
@@ -6,51 +6,35 @@
...
-				Otel: trace.Otel{
-					Enabled:  true,
-					Endpoint: "api.honeycomb.io",
-					APIKey: cfg.Secret{
-						Path: "APIKey",
-					},
-					Debug:         tracingDebug,
-					Dataset:       dataset,
-					SamplePercent: 100,
+				LogFile: trace.LogFile{
```
jaredallard, line 36: Maybe this should still be under `otel` since we're technically changing an exporter? Something like that? Not super attached

pkg/cli/logfile/logfile.go
```diff
@@ -107,9 +124,14 @@ func Hook() error {
 		<-exited
 	} else {
 		rec := newRecorder(logFile, 0, 0, cmd.Path, cmd.Args)
+		finishedChan := make(chan struct{})
```
jaredallard, line 127: Move this into the `recorder` function instead so that it can be shared between both of the functions

pkg/cli/logfile/storage_entry.go
```diff
@@ -22,12 +22,16 @@ const (
 
 	// EntryTypeFrame is a frame entry which is equal to a Frame struct
 	EntryTypeFrame
+
+	// EtnryTypeTrace is a frame entry representing a trace
```
jaredallard, line 26:
```suggestion
	// EntryTypeTrace is a frame entry representing a trace
```

pkg/cli/config.go
artemchernyak-outreach: I wanted to support this as a separate config since it's fairly different than the standard otel config. We use our own exporter, skip sampling, and generally don't allow as many configs.

pkg/cli/config.go
```diff
@@ -6,51 +6,30 @@
...
-			return nil, fmt.Errorf("failed to find secret at path '%s', or compiled into binary", path)
-		}
-
-		return fallbackSecretLookup(ctx, path)
-	})
+func overrideConfigLoaders() {
+	portStr, _ := os.LookupEnv(logfile.TracePortEnvironmentVariable)
+	port, err := strconv.Atoi(portStr)
+	if err != nil {
+		panic("invalid tracing port provided")
```
jaredallard, line 24: Maybe instead of panic drop a warning log saying traces won't be recorded?

pkg/cli/logfile/logfile.go
```diff
@@ -81,16 +89,25 @@ func Hook() error {
 	cmd.Env = os.Environ()
 	cmd.Env = append(cmd.Env, fmt.Sprintf("%s=1", EnvironmentVariable))
 
+	l, err := net.Listen(TraceSocketType, "localhost:0")
```
jaredallard, line 92: Fold the `net.Listen` and `newTraceServer` into `newRecorder`? This would allow us to duplicate less code as well as make it easier to shut everything down / wait for it to finish

pkg/cli/logfile/storage.go
```diff
@@ -62,6 +67,14 @@ type Frame struct {
 	Bytes []byte `json:"b"`
 }
 
+type Trace struct {
```
jaredallard, line 70: nit: comment at top of `Trace`

pkg/cli/logfile/storage.go
```diff
@@ -95,3 +108,78 @@ func read(r io.Reader) ([]Entry, error) {
...
+	Attributes             []attribute.KeyValue
+	Events                 []tracesdk.Event
+	Links                  []tracesdk.Link
+	Status                 tracesdk.Status
+	DroppedAttributes      int
+	DroppedEvents          int
+	DroppedLinks           int
+	ChildSpanCount         int
+	Resource               []attribute.KeyValue
+	InstrumentationLibrary instrumentation.Library
```
jaredallard, line 130: nit: Comments for each export, I know some of this is unclear so feel free to reflect that ambiguity here but just nice to have something

jaredallard: May also want to justify this duplication for future readers so they know the struggle

pkg/trace/config.go
```diff
@@ -41,6 +43,12 @@ type Otel struct {
 	AdditionalAPIKey cfg.Secret `yaml:"AdditionalAPIKey"`
 }
 
+// LogFile is the configuration for log file based tracing
+type LogFile struct {
+	//
```
jaredallard, line 48: `// Port` ? :P

pkg/trace/logfile.go
```diff
@@ -0,0 +1,147 @@
+package trace
```
jaredallard, line 1: nit: copyright + description of this file and what it's doing

pkg/trace/otel.go
```diff
@@ -37,6 +37,15 @@ type Annotator struct {
 	sampleRate int64
 }
 
+func NewOtelTracer(ctx context.Context, serviceName string, config *Config) (tracer, error) {
```
jaredallard, line 40: nit: comment for export

pkg/cli/logfile/storage.go
artemchernyak-outreach: I did add some justification at the top of the type and annotated the field that we needed to change. I feel like this represents the correct level of ambiguity since not even the OpenTelemetry library itself has any information about these fields https://pkg.go.dev/go.opentelemetry.io/otel/sdk@v1.11.0/trace/tracetest#SpanStub

pkg/cli/logfile/logfile.go
artemchernyak-outreach: I kept the listener creation here because we need it to set the environment variable and it does not result in any code duplication, but I moved all the management to the recorder.

pkg/cli/logfile/storage.go
artemchernyak-outreach: I don't feel too strongly about this, however, so I can definitely add it if you think it's necessary. I just didn't want to make descriptors for library fields that we don't own and did not define.

pkg/cli/logfile/logfile.go
jaredallard: Let me take a quick pass at it and see if it feels better, I'll let this land since it's going into my logfile PR anyways tho :D

pkg/cli/logfile/storage.go
jaredallard: Sounds reasonable!

PR #240, opened 2022-10-24
feat: add full roundtrip span to json processing

artemchernyak-outreach:

This  ensures that we can successfully encode spans to JSON and decode them back into Go structures without losing the needed data.

## Jira ID

[DT-2990]

## Notes for your reviewers

[DT-2990]: https://outreach-io.atlassian.net/browse/DT-2990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/trace/trace.go
```diff
@@ -341,11 +341,3 @@ func FromHeaders(ctx context.Context, hdrs map[string][]string, name string) con
 	}
 	return defaultTracer.fromHeaders(ctx, hdrs, name)
 }
-
-// ForceFlush immediately exports all the spans that have been buffered to this point
```
jaredallard, line 345: Breaking change?

jaredallard (approved):

pkg/trace/trace.go
artemchernyak-outreach: Actually, no. This change was introduced in the initial Delibird migration (not merged to main yet). I realized that https://github.com/getoutreach/gobox/pull/240/files#diff-5b8f2be4e453643719aa5e585f8e8a1d73bfb8b873c1ac8d3902279f1b2e90f5R50 will force a flush for us.

PR #243, opened 2022-11-08
feat: intro trace.IsSampled method

nissim-natanov-outreach:
This PR introduces IsSampled method to avoid prevention of 'dead links' in linking scenarios. E.g.:
* clerk producer attaches trace headers to clerk message
* clerk consumer reads trace header

Without the IsSampled: it always links to this header, but in Honeycomb this link will be DEAD. It is virtually impossible to show traces with link which are NOT dead - thus to find a trace with a valid link requires at least 100 clicks

With IsSampled: I will patch clerk code to only attach link if the source was sampled. Otherwise, it will be ignored. This way when looking at honeycomb, we can now check the traces with Links and those trace links will likely be valid ones.

Usage here: https://github.com/getoutreach/clerk/pull/344

## Jira ID

[CDT-203]

## Notes for your reviewers

[CDT-203]: https://outreach-io.atlassian.net/browse/CDT-203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/trace/force_sampler.go
+++ b/pkg/trace/force_sampler.go
@@ -38,11 +38,6 @@ func (s *otelForceSampler) ShouldSample(p sdktrace.SamplingParameters) sdktrace.
 		}
 	}
 
-	sampler, err := sample.NewDeterministicSampler(s.sampleRate)
-	if err != nil {
-		panic(err)
-	}
-
 	var forceTrace string
 	for _, a := range p.Attributes {
 		if string(a.Key) == fieldForceTrace {
@@ -57,8 +52,8 @@ func (s *otelForceSampler) ShouldSample(p sdktrace.SamplingParameters) sdktrace.
 		}
 	}
 
-	traceID := p.TraceID.String()
-	if sampler.Sample(traceID) {
+	// if not forced, use deterministic hash of the trace ID and the current rate to decide
+	if s.isSampled(p.TraceID.String()) {
 		return sdktrace.SamplingResult{
 			Decision:   sdktrace.RecordAndSample,
 			Tracestate: psc.TraceState(),
@@ -71,8 +66,27 @@ func (s *otelForceSampler) ShouldSample(p sdktrace.SamplingParameters) sdktrace.
 	}
 }
 
-func forceSample(sampleRate uint) sdktrace.Sampler {
+// isSampled is used to determine if the given trace ID should be sampled based on the
+// current sample rate.
+//
+// Please keep the sampling logic deterministic because this method is used to correlate sampling between different
+// services at Outreach. For example, if a producer of the trace headers generates traceID1 and a remote service
+// wants to link the traceID1 to its own trace (!= traceID1), then it should only do so if traceID1 was sampled.
+// Failure to do so causes honeycomb to show a trace link to a 'dead trace' - and it is not possible to filter those
+// out, rendering the the cross-trace linking experience useless.
+func (s *otelForceSampler) isSampled(traceID string) bool {
+	sampler, err := sample.NewDeterministicSampler(s.sampleRate)
+	if err != nil {
+		panic(err)
+	}
+
+	return sampler.Sample(traceID)
+}
+
+func newSampler(sampleRate uint) *otelForceSampler {
 	return &otelForceSampler{
 		sampleRate: sampleRate,
 	}
 }
+
+var _ sdktrace.Sampler = &otelForceSampler{}
--- a/pkg/trace/force_sampler.go
+++ b/pkg/trace/force_sampler.go
@@ -21,7 +21,7 @@ func forceTracing(ctx context.Context) context.Context {
 // forceSampler allows force sample rate to 100% when trace context contains field force_trace
 // and sample at a rate of 1/<given rate> when context contains field sample_trace.
 type otelForceSampler struct {
-	sampleRate uint
+	sampler *sample.DeterministicSampler
 }
 
 func (s *otelForceSampler) Description() string {
@@ -75,17 +75,17 @@ func (s *otelForceSampler) ShouldSample(p sdktrace.SamplingParameters) sdktrace.
 // Failure to do so causes honeycomb to show a trace link to a 'dead trace' - and it is not possible to filter those
 // out, rendering the the cross-trace linking experience useless.
 func (s *otelForceSampler) isSampled(traceID string) bool {
-	sampler, err := sample.NewDeterministicSampler(s.sampleRate)
+	return s.sampler.Sample(traceID)
+}
+
+func newSampler(sampleRate uint) *otelForceSampler {
+	sampler, err := sample.NewDeterministicSampler(sampleRate)
 	if err != nil {
 		panic(err)
 	}
 
-	return sampler.Sample(traceID)
-}
-
-func newSampler(sampleRate uint) *otelForceSampler {
 	return &otelForceSampler{
-		sampleRate: sampleRate,
+		sampler: sampler,
 	}
 }
 
--- a/pkg/trace/force_sampler.go
+++ b/pkg/trace/force_sampler.go
@@ -78,6 +78,7 @@ func (s *otelForceSampler) isSampled(traceID string) bool {
 	return s.sampler.Sample(traceID)
 }
 
+// newSampler creates a new deterministic sampler for the given rate or panics if the rate is 0.
 func newSampler(sampleRate uint) *otelForceSampler {
 	sampler, err := sample.NewDeterministicSampler(sampleRate)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


artemchernyak-outreach (approved):
Looks great. Good idea exposing this as it will allow us to more easily support all the options through the kafka headers as well as other types.

pkg/trace/force_sampler.go
```diff
@@ -71,8 +66,27 @@ func (s *otelForceSampler) ShouldSample(p sdktrace.SamplingParameters) sdktrace.
...
+// isSampled is used to determine if the given trace ID should be sampled based on the
+// current sample rate.
+//
+// Please keep the sampling logic deterministic because this method is used to correlate sampling between different
+// services at Outreach. For example, if a producer of the trace headers generates traceID1 and a remote service
+// wants to link the traceID1 to its own trace (!= traceID1), then it should only do so if traceID1 was sampled.
+// Failure to do so causes honeycomb to show a trace link to a 'dead trace' - and it is not possible to filter those
+// out, rendering the the cross-trace linking experience useless.
+func (s *otelForceSampler) isSampled(traceID string) bool {
+	sampler, err := sample.NewDeterministicSampler(s.sampleRate)
```
vk-outreach, line 78: While it is true that `sample.NewDeterministicSampler` is stateless, we should probably still not allocate this on each sample call.  I know the earlier code used to do the same but can we take the chance to fix this by adding the sampler as state to otelForceSampler struct (as part of the newSampler call).

Also, I'd move the panic to the `newSampler` call rather than have panic in the `shouldSample` code path.

pkg/trace/trace.go
```diff
@@ -323,3 +323,12 @@ func FromHeaders(ctx context.Context, hdrs map[string][]string, name string) con
 	}
 	return defaultTracer.fromHeaders(ctx, hdrs, name)
 }
+
+// IsSampled returns true if the given trace ID is sampled based on the current sampling rate.
+// This API is mostly for remote trace ID correlation.
+func IsSampled(traceID string) bool {
```
vk-outreach, line 329: Just curious: do we need `traceID` as a param here?  Wouldn't it be better to make this API take the `context` and identify if the current trace is being sampled or not?

pkg/trace/force_sampler.go
jaredallard: +1 here

pkg/trace/tracer.go
```diff
@@ -54,4 +54,6 @@ type tracer interface {
 
 	// fromHeaders is similar to contextFromHeaders + it starts a new span
 	fromHeaders(ctx context.Context, hdrs map[string][]string, name string) context.Context
+
```
jaredallard, line 57: nit: function comments please!

jaredallard: (applies to all functions in this PR)

pkg/trace/force_sampler.go
nissim-natanov-outreach: sure, I have moved the init logic with the panic into the newSampler in the last iteration

pkg/trace/trace.go
nissim-natanov-outreach: We have the trace.ID()` function which returns the trace ID out of the context ... so this part is covered.

As for the IsSampledWithContext - my wild guess is it is not needed because it can be easily achieved with `trace.IsSampled(trace.ID())`

I do not have a strong opinion around either using the context (and hiding trace ID as impl) - but given that we already have the trace.ID function to expose raw traceID out it would be awkward to ask people to wrap the trace ID back into the context in order to get the IsSampled method called on it (in case they already pass the ID as a string around)

vk-outreach: If we support enabling/disabling sampling or changing sampling rate based on stuff in the context, an API that is based on context is likely to be more accurate.

Not disputing whether we need the traceID based API as well but I think if the intention is to figure if the current trace is sampled, it would be better to model that API via the context.

nissim-natanov-outreach: The intention of this API is to figure out if the remote trace ID is sampled, not the local one, so that it could be 'safely linked' by the recipient of the trace.

It is very hard to think of a case where somebody would need IsSampled equivalent for the local traces. If yes, then we can always add dedicated IsSampledFromContext that will take care of trace ID + whatever local overrides where applied to force the sample.

vk-outreach: An alternate design is to model it like so:

```golang
    trace.CurrentSampler(ctx).IsSampled(trace.ID(ctx)) ...
```

It does require a lot more changes but it is probably more elegant as we can make the `trace.CurrentSampler` return the default sampler for the offline cases.

pkg/trace/tracer.go
nissim-natanov-outreach: done

pkg/trace/trace.go
vk-outreach: If the API is explicitly for remote traces, how would we deal with sample rates and such?

It seems wrong to use the current sampling configuration for remote traces and so I'd suggest we do something like `trace.Sampler(config/sample_rate).IsSampled(traceID)` for that case rather than `trace.IsSampled`.

nissim-natanov-outreach: Actually, Link does not expose raw Context object - simply because it is a Link and we do not want to accidentally generate context out of the link. That means the link itself should be the one to tell if it is sampled or not.

Maybe IsSampled should be moved into the Link struct instead to clearly scope it to the link only?

nissim-natanov-outreach: I moved it to the Link struct to scope it down, and renamed into IsSampledIfLocal + doc about heuristics and reasoning around this. LMK if it looks better.

artemchernyak-outreach: Actually there is another concern here. Although we are sampling at the same rate everywhere today, that will not always be the case. For example is service A is sampling at 10% and service B is sampling 1%, then service B would still have a lot of misses for the traces from service A.

To solve this, I think we have two options:

- We need to pass through the original sampling percent and use that to create a new sampler at the same percentile. This seems like a lot to do in order to link the traces.
- We can pass through a force header only on the traces that get sampled from A. That would mean that regardless of the sampling rate of B, we will force B to sample that trace.

artemchernyak-outreach:
We should also definitely write some tests for this. Would be nice to cover at least these three scenarios for sampling successfully:

- Samplers with the same sampling rate 
- The produce sampler having a lower rate than the consumer sampler
- The consumer sampler having a lower rate than the producer sampler

Ideally, all of the above cases should sample as long as the producer samples.

pkg/trace/trace.go
nissim-natanov-outreach: this is indeed a better fix, but I propose we do it incrementally because it wont break API. Without the current fix e2e trace feature is completely unusable - so with this fix we will at least have the MOST COMMON case covered in which sampling between producer and consumer are same. With your fix it will be even more accurate indeed, but it requires three more iterations that IMHO can happen later too.

vk-outreach: The earlier suggestion I had (see below) forces the sample rate to be explicit -- this will atleast make it clear that code built with it is going to have all these problems.   I'd personally prefer we do the force-trace approach that @artemchernyak-outreach suggests though -- better to have things be robust when it comes to observability.

> It seems wrong to use the current sampling configuration for remote traces and so I'd suggest we do something like trace.DefaultSampler(config/sample_rate).IsSampled(traceID) for that case rather than trace.IsSampled.

nissim-natanov-outreach:
> We should also definitely write some tests for this. Would be nice to cover at least these three scenarios for sampling successfully:
> 
> * Samplers with the same sampling rate
> * The produce sampler having a lower rate than the consumer sampler
> * The consumer sampler having a lower rate than the producer sampler
> 
> Ideally, all of the above cases should sample as long as the producer samples.

I have filed Jira [DT-3105] for the DT to complete the picture with a new header and consider new e2e or other tests (maybe with Ruby-equivalent algo around to check that the sampling between the two is actually deterministic).

pkg/trace/trace.go
nissim-natanov-outreach: @vk-outreach clerk library does not have access to the inner sample rate - it is deep inside the trace configuration. 

If you would still prefer to expose the Sampler, maybe it should be trace.CurrentSampler() instead which uses the hidden sample rate instead of forcing callers to magically  get the sample rate somehow.

nissim-natanov-outreach:
I chatted with VK offline and we have agreed to avoid changes in gobox and instead temporary clone the sampler logic into clerk itself, e.g. https://github.com/getoutreach/clerk/pull/344/files#diff-786ba6959806e8385c644e71dd201d17112372ea8a86d5bf15f57a5606e44574R40. I am cancelling this PR.

We will wait for DT to implement the proposed solution and then remove clerk's hack.

PR #247, opened 2022-11-14
DRAFT: Proper types for OTel span attrs

richardlarocque:
A proof-of-concept that attempts to use the right attrbute type for the given input.  As of right now, everything is being reported up as a string, and HC sees it as a string, which means we don't get support for `SUM` or `AVG` or other numeric-only functions on numeric data.

We can always fix this manually in the UX, but ideally we shouldn't need to.

### DIFF
--- a/pkg/trace/otel.go
+++ b/pkg/trace/otel.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/getoutreach/gobox/internal/logf"
 	"github.com/getoutreach/gobox/pkg/app"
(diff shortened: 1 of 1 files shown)


pkg/trace/otel.go
```diff
@@ -232,7 +233,43 @@ func (t *otelTracer) addInfo(ctx context.Context, args ...log.Marshaler) {
...
+				case float64:
+					span.SetAttributes(attribute.Float64(key, v))
+				case string:
+					span.SetAttributes(attribute.String(key, v))
+				case time.Time:
+					// This is a compromise.  OTel seems to
+					// prefer UNIX epoch milliseconds, while
+					// Honeycomb says it accepts UNIX epoch
+					// seconds.  Honeycomb also has a function to
+					// convert RFC3339 timestmaps to epoch.
```
jkinkead, line 262:
```suggestion
					// convert RFC3339 timestamps to epoch.
```

jkinkead (commented):
One typo.

I'm not familiar enough to approve. FYI @artemchernyak-outreach - if you haven't seen this.

richardlarocque:
I'm abandoning this in favor of https://github.com/getoutreach/gobox/pull/248.

PR #249, opened 2022-11-19
feat: change queue capacity type to uint

LophyLiu:
It does not make sense for a queue capacity to be negative. This PR changes the queue capacity type to be `uint` so that creating a queue is simpler without handling the return error. 

## Jira ID

[XX-XX]

### DIFF
--- a/pkg/queue/example_queue_test.go
+++ b/pkg/queue/example_queue_test.go
@@ -11,10 +11,7 @@ import (
 
 func ExamplePriorityQueue() {
 	// Create priority queue.
-	pq, err := queue.NewPriorityQueue()
-	if err != nil {
-		panic(err)
-	}
+	pq := queue.NewPriorityQueue()
 
 	// Push items.
 	pushItem := func(data interface{}, priority int64) *queue.PriorityQueueItem {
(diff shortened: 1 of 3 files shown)


jkinkead (changes requested):
This makes some sense, but it's a breaking change (and would preclude adding other error checks to the constructor).

If this were non-breaking (a new constructor + new option), then we could roll this out, update callers, and deprecate the old function.

I'm _still_ not sure it's worth removing the possibility of adding more error checks to the constructor in the future, though. :)

clevelittlefield-outreach:
> This makes some sense, but it's a breaking change (and would preclude adding other error checks to the constructor).
> 
> If this were non-breaking (a new constructor + new option), then we could roll this out, update callers, and deprecate the old function.
> 
> I'm _still_ not sure it's worth removing the possibility of adding more error checks to the constructor in the future, though. :)

I am pretty sure we are the only people using this... Usually our contructors dont have errors, so by switching to uint we are able to go back to a natural state

jkinkead (approved):
Happy to approve if the only customers of this are happy with the required changes. :)

PR #250, opened 2022-11-21
fix: add cd to box

chatchai-outreach:

expose cd for devbase to be able to get maestro url via box instead of hardcoding 

## Jira ID

https://outreach-io.atlassian.net/browse/DT-2953


tonymbugua-outreach (approved):
LGTM - adds CD

PR #251, opened 2022-11-21
fix: env test override data race

george-e-shaw-iv:

* env.FakeTestConfig was causing a data race in parallelized tests
  because it wasn't protected via a mutex (or anything to prevent a
  data race).

## Jira ID

[XX-XX]

### DIFF
--- a/pkg/env/readers.go
+++ b/pkg/env/readers.go
@@ -12,14 +12,52 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
+	"sync"
 
 	"github.com/getoutreach/gobox/pkg/app"
 	"github.com/getoutreach/gobox/pkg/cfg"
 	"gopkg.in/yaml.v3"
 )
 
+type testOverrides struct {
+	data map[string]interface{}
+	mu   sync.Mutex
+}
+
+func (to *testOverrides) add(k string, v interface{}) {
+	to.mu.Lock()
+	defer to.mu.Unlock()
+
+	to.data[k] = v
+}
+
+func (to *testOverrides) load(k string) (interface{}, bool) {
+	to.mu.Lock()
+	defer to.mu.Unlock()
+
+	// Apparently you cannot pull the bool out of this access implicitly in the return
+	// statement.
+	v, ok := to.data[k]
+
+	return v, ok
+}
+
+func (to *testOverrides) delete(k string) {
+	to.mu.Lock()
+	defer to.mu.Unlock()
+
+	delete(to.data, k)
+}
+
 // nolint:gochecknoglobals // Why: needs to be overridable
-var testOverrides = make(map[string]interface{})
+var overrides testOverrides
+
+//nolint:gochecknoinits // Why: Tech debt on testOverrides being a thread-unsafe global (used to just be a map w/no mutex).
+func init() {
+	overrides = testOverrides{
+		data: make(map[string]interface{}),
+	}
+}
 
 // linter is not aware of or_dev tags, so it falsely considers this deadcode.
 func devReader(fallback cfg.Reader) cfg.Reader { //nolint:deadcode,unused // Why: only used with certain build tags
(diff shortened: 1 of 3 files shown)


pkg/env/readers.go
```diff
@@ -58,9 +96,9 @@ func devReader(fallback cfg.Reader) cfg.Reader { //nolint:deadcode,unused // Why
 	})
 }
 
-func testReader(fallback cfg.Reader, overrides map[string]interface{}) cfg.Reader {
+func testReader(fallback cfg.Reader, overrides *testOverrides) cfg.Reader {
```
jkinkead, line 99: This function parameter name shadows the module-level variable, which is pretty bad for readability. Rename one of them?

pkg/env/readers.go
```diff
@@ -12,14 +12,52 @@ import (
...
+	defer to.mu.Unlock()
+
+	delete(to.data, k)
+}
+
 // nolint:gochecknoglobals // Why: needs to be overridable
-var testOverrides = make(map[string]interface{})
+var overrides testOverrides
+
+//nolint:gochecknoinits // Why: Tech debt on testOverrides being a thread-unsafe global (used to just be a map w/no mutex).
```
jkinkead, line 55: Why can't you initialize inline above?

pkg/env/readers.go
```diff
@@ -71,16 +109,16 @@ func testReader(fallback cfg.Reader, overrides map[string]interface{}) cfg.Reade
...
+	if _, ok := overrides.load(fName); ok {
 		// This is not ideal.  We would prefer to return an error.
 		// However, this function's signature does not support it and we
 		// don't want to incur the backwards-incompatibility of changing
 		// it right now.
 		panic(fmt.Errorf("repeated test override of '%s'", fName))
 	}
 
-	testOverrides[fName] = ptr
+	overrides.add(fName, ptr)
```
jkinkead, line 120: This still has a race condition, because you're not copying things. You need a per-goroutine _copy_ of the structure, else you're still sharing memory.

pkg/env/readers.go
```diff
@@ -12,14 +12,52 @@ import (
...
 	"os/user"
 	"path/filepath"
+	"sync"
 
 	"github.com/getoutreach/gobox/pkg/app"
 	"github.com/getoutreach/gobox/pkg/cfg"
 	"gopkg.in/yaml.v3"
 )
 
+type testOverrides struct {
```
jaredallard, line 22: Not seeing the value of this over `sync.Map` at this point. Thoughts on using that?

jkinkead: Also does not fix the race condition of two tests adding to that map at the same time.

george-e-shaw-iv: sync.Map isn't structured and they actually warn you against using it in the godoc for it unless its for very specific things.

jaredallard: Fine w/ me, just seemed super duplicated. I get the type requirements here, but w/ just strings it didn't seem that beneficial

pkg/env/readers.go
george-e-shaw-iv: changing it to overrider

pkg/env/readers.go
george-e-shaw-iv: good call

pkg/env/readers.go
```diff
@@ -71,16 +112,11 @@ func testReader(fallback cfg.Reader, overrides map[string]interface{}) cfg.Reade
...
 // The provided value is serialized to yaml and so can be structured data.
 func FakeTestConfig(fName string, ptr interface{}) func() {
-	if _, ok := testOverrides[fName]; ok {
-		// This is not ideal.  We would prefer to return an error.
-		// However, this function's signature does not support it and we
-		// don't want to incur the backwards-incompatibility of changing
-		// it right now.
-		panic(fmt.Errorf("repeated test override of '%s'", fName))
-	}
+	// add ensures that it doesn't already exist to prevent two tests running
```
jkinkead, line 115: Document panic in function signature, and document potential leakage to other tests?

This still leaks shared state: You can read overrides of _other_ functions, even if you're only mocking out one.

This whole thing is a nightmare, and honestly the right call might be to make a breaking change (or create a correct version clients could migrate to).

At the bare minimum this needs a TODO + Jira bug.

george-e-shaw-iv: Okay, I'm going to create a jira bug because I was just kinda doing this one out of kindness - I was originally just going to open a DT bug but decided to try and help out. I can't really spend anymore time on this with other commitments right now - this is kind of a tight sprint w/the holidays.

george-e-shaw-iv: https://outreach-io.atlassian.net/browse/DT-3185

jkinkead (approved):
LGTM! Thanks for pitching in.

jaredallard (approved):

PR #254, opened 2022-12-14
fix: deprecate runbackground

george-e-shaw-iv:

* runbackground is dangerous. you should always be aware of and control your context.

## Jira ID

[STAR-759]

## Notes for your reviewers

[STAR-759]: https://outreach-io.atlassian.net/browse/STAR-759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/async/async.go
+++ b/pkg/async/async.go
@@ -121,6 +121,8 @@ func Run(ctx context.Context, r Runner) {
 // RunBackground executes a single asynchronous task with background context
 //
 // It creates a new trace for the task and passes through deadlines.
+//
+// deprecated: Please only use Run - you should always be aware of and control your context.
 func RunBackground(ctx context.Context, r Runner) {
 	Run(context.Background(), r)
 }
--- a/pkg/async/async.go
+++ b/pkg/async/async.go
@@ -122,7 +122,7 @@ func Run(ctx context.Context, r Runner) {
 //
 // It creates a new trace for the task and passes through deadlines.
 //
-// deprecated: Please only use Run - you should always be aware of and control your context.
+// Deprecated: Please only use Run - you should always be aware of and control your context.
 func RunBackground(ctx context.Context, r Runner) {
 	Run(context.Background(), r)
 }


pkg/async/async.go
```diff
@@ -121,6 +121,8 @@ func Run(ctx context.Context, r Runner) {
 // RunBackground executes a single asynchronous task with background context
 //
 // It creates a new trace for the task and passes through deadlines.
+//
+// deprecated: Please only use Run - you should always be aware of and control your context.
```
george-e-shaw-iv, line 125:
```suggestion
// Deprecated: Please only use Run - you should always be aware of and control your context.
```

zaquestion (changes requested):
blocking, to allow time to remove this from orgservice generation

jaredallard:
> blocking, to allow time to remove this from orgservice generation

Why? If you don't upgrade `gobox` you won't get the deprecation warning/lint

pkg/async/async.go
```diff
@@ -121,6 +121,8 @@ func Run(ctx context.Context, r Runner) {
 // RunBackground executes a single asynchronous task with background context
 //
 // It creates a new trace for the task and passes through deadlines.
+//
+// Deprecated: Please only use Run - you should always be aware of and control your context.
```
jaredallard, line 125:
```suggestion
// Deprecated: Use Run with context.Background() instead (if required).
```

jaredallard (approved):

zaquestion:
I'm more concerned about all the integrations that may pull upgraded gobox and be unable to push because of code they can't control

pkg/async/async.go
zaquestion:
```suggestion
// Deprecated: Use Run with an explicit context.WithTimeout to prevent goroutine leak
```
Fine with the current wording, but if any thing I'd change to more like

zaquestion (approved):
evidently it wasn't in the templates https://github.com/getoutreach/orgservice/pull/676

PR #256, opened 2023-01-05
fix(cli): try load trace.yaml before overriding

AndrewWinterman:

when run in production, the trace.yaml override breaks tracing for
containers that don't use ldflags to set the honeycomb api key.

This simply tries to load trace.yaml before falling back to the CLI's
overrides

## Jira ID

[QSS-1588]

## Notes for your reviewers

[QSS-1588]: https://outreach-io.atlassian.net/browse/QSS-1588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/cli/bootstrap.go
+++ b/pkg/cli/bootstrap.go
@@ -58,8 +58,12 @@ func overrideConfigLoaders(honeycombAPIKey, dataset string, tracingDebug bool) {
 	})
 
 	fallbackConfigReader := cfg.DefaultReader()
+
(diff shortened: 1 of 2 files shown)


artemchernyak-outreach (approved):
Seems like a great solution as we run more clis inside deployments.

jaredallard (approved):

PR #258, opened 2023-01-11
feat(trace): set error status when there are errors

AndrewWinterman:
Set otel status; add trace sugar

Errors are annoying right now.

You have to convert an `error` into a log.Marshaler, usually through events.ErrorInfo, (which isn't itself an error, so you have to return your original error from your function), and then, for your trouble, you don't even get the error status set correctly on your span.

I am aware the call framework exists, and likely does nice things on the span when used, but for library authors (e.g. QSS & smartstore) it doesn't make sense to either assume we are in or start a call, most of the time, but we still want to know when our infrastructure errors in traces.

This commit adds an Error convenience function, and special handling in otel.go for both `error` and `events.ErrorInfo` objects so the status is set. This should improve the utility of honeycomb, and other tools that assume you'll use the OTEL spec's `code` field. The commit should be considered a proposal until discussion with tooling is resolved, and tests added.

## Jira ID

[XX-XX]


pkg/events/error.go
```diff
@@ -27,6 +27,21 @@ type ErrorInfo struct {
 	Custom   log.Marshaler
 }
 
+// LoggableError is the minimum required to make an error loggable.
```
samnguyen-outreach, line 30: This seems less powerful than the existing [events.NewError](https://pkg.go.dev/github.com/getoutreach/gobox@v1.57.3/pkg/events#NewErrorInfo), which already implements `log.Marshaler`.  Could that be used instead?

AndrewWinterman: TBH this is overcomplicated, I just wanted a thing that would let me pass an error to status, with minimal fuss. I ended up moving `loggableError` to the trace package where it only does that.

I have no real objection to using events.NewError instead, except that I worry it will make the status in traces hard to read due to its length, and if you use the new trace.Error fn, it will duplicate information available at other points int he span.

pkg/trace/trace.go
```diff
@@ -255,6 +255,25 @@ func AddInfo(ctx context.Context, args ...log.Marshaler) {
 	}
 }
 
+// Error is a convenience for attaching an error to a span.
+func Error(ctx context.Context, err error) error {
```
richardlarocque, line 259: Why not make this simply `AddInfo(ctx, events.NewErrorInfo(err))`?

AndrewWinterman: Likely a distinct concern, but because I'm not sure what to make of traces with fields like: https://ui.honeycomb.io/outreach-a0/datasets/outreach/result/3fhi6d8M9A6?tab=traces

My thought was to provide something dead simple. Maybe it's worse though? Seems like it might be better to simply stick to the existing types and lead the loggableError type out.

richardlarocque: Yeah, it is a bit much, but I'm not sure there's a great alternative.  Sometimes the key to understanding what's wrong actually is buried 5 causes deep.

AndrewWinterman: I think regardless, it's an unrelated issue, so I'm just going to use the existing ErrorInfo type and try to formulate some more precise thoughts on how the error gets serialized.

The last thing I'll say about it here is I think this format makes some important questions more difficult to answer, like what is the error distribution for my service?

pkg/events/error.go
AndrewWinterman: Fixed

pkg/trace/trace.go
AndrewWinterman: Fixed in [2187936](https://github.com/getoutreach/gobox/pull/258/commits/21879361059969b58cecc80dd9d010320e201bb5)

pkg/trace/otel.go
```diff
@@ -224,6 +225,17 @@ func (t *otelTracer) addInfo(ctx context.Context, args ...log.Marshaler) {
 		for _, arg := range args {
 			kvs := marshalToKeyValue(arg)
 			span.SetAttributes(kvs...)
+
+			// setting the code will cause error spans to be called out specifically
```
richardlarocque, line 229: I was digging into the docs on this `SetStatus` thingy and I found this: https://opentelemetry.io/docs/instrumentation/go/manual/#record-errors.

Should we be doing `span.SetError(v.Error)` here?

pkg/trace/otel.go
```diff
@@ -224,6 +225,17 @@ func (t *otelTracer) addInfo(ctx context.Context, args ...log.Marshaler) {
...
+			// in the trace, and will make the "errors" default view in honeycomb useful
+			switch v := arg.(type) {
+			case *events.ErrorInfo:
+				span.SetStatus(codes.Error, v.Error)
+			case error:
+				span.SetStatus(codes.Error, v.Error())
```
richardlarocque, line 235: `error` doesn't meet the requirements of `log.Marshaler`, right?  Therefore `arg.(type)` will never be `error` and we should not hit this branch?

AndrewWinterman: maybe a comment is in order-- nothing stops clients from defining an `error` that implements `log.Marshaler`, so this is trying to catch that case, since i think in that circumstance it's reasonable for the client to expect the error to be respected.

pkg/trace/otel.go
AndrewWinterman: I think you mean both

```
	span.SetStatus(codes.Error, v.Error)
	span.RecordError(v)
```

I think yes? that does seem better. IIRC, and it's been a while since i experimented with this, RecordError will just set the error attribute in the span, which might clober what's done above, which is probably fine?

AndrewWinterman: hokay, i looked into this-- it adds an event, seems reasonable to me https://github.com/open-telemetry/opentelemetry-go/blob/9f42a81a5f8b4faf5689b72f9e4093361d8e5885/sdk/trace/span.go#L369

richardlarocque (approved):
LGTM.

PR #259, opened 2023-01-11
fix(app_domain): add domain to app.Info

surendra-outreach:
This change adds app domain info to the app.Info() data. We are doing this change to add domain info to the mint tokens which will used (for example) in [authnidentity/userprofile](https://github.com/getoutreach/authnidentity/blob/f58edaa7caede96c88f18000c17ea4f474856c73/internal/handlers/identity.go#L114) to decode the user token in the context. 

## Jira ID

[IATS-754](https://outreach-io.atlassian.net/browse/IATS-754)

## Notes for your reviewers

[IATS-754]: https://outreach-io.atlassian.net/browse/IATS-754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -96,6 +96,16 @@ func info() *Data {
 		environment = env
 	}
 
+	var domain string
+	switch environment {
+	case "development":
+		domain = "outreach-dev.com"
+	case "staging":
+		domain = "outreach-staging.com"
+	default: // production
+		domain = "outreach.io"
+	}
+
 	clusterName := unknown
 	if cn := os.Getenv("MY_CLUSTER"); cn != "" {
 		clusterName = cn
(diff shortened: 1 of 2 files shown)


pkg/app/app.go
```diff
@@ -96,6 +96,16 @@ func info() *Data {
 		environment = env
 	}
 
+	var domain string
+	switch environment {
+	case "development":
```
jaredallard, line 101: This is open source, so storing these here doesn't really work. I'd suggest a different package that consumes `app.Info()` and returns these values as required.

jaredallard: `ServiceID`, realistically, shouldn't have been put here either.

My other concern is also with mixing authn and application information, which I think should be owned entirely by an authn specific package only (as I somewhat implied above)

surendra-outreach: ok, in that case, i will move this into mint, and will close this pr. thx.

PR #260, opened 2023-02-01
feat: deprecate  testOverrides to enable better race condition handling 

tonymbugua-outreach:
The current function signature for `testOverrides` does not handle race conditions well and will throw a panic instead of returning an error. This PR will deprecate the `testOverrides` struct and create a new struct that allows for an error to be returned as a function signiture.

## Jira ID

DT-3185


pkg/env/readers.go
```diff
@@ -126,3 +138,16 @@ func FakeTestConfig(fName string, ptr interface{}) func() {
 		overrides.delete(fName)
 	}
 }
+
+// FakeTestConfigHandler allows you to fake the test config with a specific value.
+// and returns an error if a config with the same name exists already
```
jkinkead, line 143: Period on wrong line. I would also expand very slightly to note how to handle errors:
```suggestion
// FakeTestConfigHandler allows you to fake the test config with a specific value
// and returns an error if a config with the same name exists already. If callers get an error,
// they should switch to running tests in serial.
```

pkg/env/readers.go
```diff
@@ -116,7 +128,7 @@ func testReader(fallback cfg.Reader, overrider *testOverrides) cfg.Reader {
 // use the fName across two tests running in parallel. This will cause the
 // function to potentially panic.
 //
-// TODO[DT-3185]: Related work item to make the safety of this function better
+// Deprecated: Please use `FakeTestConfigHandler`
```
jkinkead, line 131: You should replace the call on line 135 with a call to the new `addHandler` function - and delete the old `add` function.

This function can now have the `panic` statement, and you can have the note about this being bad & deprecated.

pkg/env/readers_test.go
```diff
@@ -0,0 +1,148 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
+
+//go:build or_dev || or_test || or_e2e
+// +build or_dev or_test or_e2e
+
+// Description: Provides configuration readers for various environments
```
jkinkead, line 6: Is this even required by the linter?

If so, you should describe the file contents:
```suggestion
// Description: Unit tests for configuration readers
```

pkg/env/readers_test.go
```diff
@@ -0,0 +1,148 @@
...
+// are created with different names
+func TestFakeTestConfigHandlerMultipleConfigs(t *testing.T) {
+	type args struct {
+		fName  string
+		config TestConfig
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    func()
```
jkinkead, line 47: Unread struct field. Remove?

pkg/env/readers_test.go
```diff
@@ -0,0 +1,148 @@
...
+			args: args{
+				fName: "test1.yaml",
+				config: TestConfig{
+					ListenHost: "another-url",
+					HTTPPort:   8000,
+					GRPCPort:   9000,
+				},
+			},
+			want:    func() {},
+			wantErr: false,
```
jkinkead, line 61: Neither fixture has this `true`. Add an error case, or remove the field?

pkg/env/readers_test.go
```diff
@@ -0,0 +1,148 @@
...
+// with the same name are created
+func TestFakeTestConfigHandlerRepeatedTestOverride(t *testing.T) {
+	type args struct {
+		fName  string
+		config TestConfig
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    func()
```
jkinkead, line 106: Unread struct field, remove.

jkinkead (commented):
A couple test case suggestions.

pkg/env/readers.go
```diff
@@ -116,13 +128,29 @@ func testReader(fallback cfg.Reader, overrider *testOverrides) cfg.Reader {
...
 func FakeTestConfig(fName string, ptr interface{}) func() {
 	// add ensures that it doesn't already exist to prevent two tests running
 	// concurrently colliding on fName.
-	overrides.add(fName, ptr)
-
+	err := overrides.addHandler(fName, ptr)
+	if err != nil {
```
jaredallard, line 136:
```suggestion
	if err := overrides.addHandler(fName, ptr); err != nil {
```

pkg/env/readers.go
```diff
@@ -116,13 +128,29 @@ func testReader(fallback cfg.Reader, overrider *testOverrides) cfg.Reader {
...
 		overrides.delete(fName)
 	}
 }
+
+// FakeTestConfigHandler allows you to fake the test config with a specific value
+// and returns an error if a config with the same name exists already. If callers get an error,
+// they should switch to running tests in serial.
+func FakeTestConfigHandler(fName string, ptr interface{}) (func(), error) {
+	err := overrides.addHandler(fName, ptr)
+	if err != nil {
```
jaredallard, line 149:
```suggestion
	if err := overrides.addHandler(fName, ptr); err != nil {
```

pkg/env/readers.go
```diff
@@ -24,19 +24,31 @@ type testOverrides struct {
...
+// 	if _, exists := to.data[k]; exists {
+// 		// This is not ideal.  We would prefer to return an error. However
+// 		// the caller function's signature does not support it and we don't
+// 		// want to incur the backwards-incompatibility of changing it right
+// 		// now.
+// 		panic(fmt.Errorf("repeated test override of '%s'", k))
+// 	}
+
+// 	to.data[k] = v
+// }
```
jaredallard, line 40: Delete code instead of commenting it out -- we have it in git history!

pkg/env/readers.go
```diff
@@ -116,13 +128,29 @@ func testReader(fallback cfg.Reader, overrider *testOverrides) cfg.Reader {
...
+	}
 	return func() {
 		overrides.delete(fName)
 	}
 }
+
+// FakeTestConfigHandler allows you to fake the test config with a specific value
+// and returns an error if a config with the same name exists already. If callers get an error,
+// they should switch to running tests in serial.
+func FakeTestConfigHandler(fName string, ptr interface{}) (func(), error) {
```
jaredallard, line 131: In Go it's more idiomatic, when renaming a function signature, to rename it to indicate what has changed. For example, when adding a Context you'd do `func <Name>Context`.

With that in mind, I'd suggest naming this `FakeTestConfigWithError` or something along those lines. Otherwise, it's not very clear to the consumer what we're doing here.

We should also change `FakeTestConfig` to do _nothing_ but call `FakeTestConfigHandler` (as it is named today) instead of duplicating the logic.

pkg/env/readers.go
```diff
@@ -116,13 +113,27 @@ func testReader(fallback cfg.Reader, overrider *testOverrides) cfg.Reader {
 // use the fName across two tests running in parallel. This will cause the
```
jaredallard, line 113: Update this comment as well. It should no longer panic

jaredallard (approved):

chatchai-outreach (approved):

PR #267, opened 2023-02-27
feat: Adding "module" field to logs with module of actual logger

deregtd:
Walking stack with caching stack frame lookups to find out who is calling the logger and adding it to all formatted logs.  Intent is to be able to filter out what application is the actual source of logs for filtering/cost attribution.

### DIFF
--- a/pkg/callerinfo/callerinfo.go
+++ b/pkg/callerinfo/callerinfo.go
@@ -9,14 +9,35 @@ package callerinfo
 import (
 	"fmt"
 	"runtime"
+	"runtime/debug"
+	"strings"
+	"sync"
 )
 
-var moduleLookupByPC = make(map[uintptr]string)
+type CallerInfo struct {
+	Function      string
+	File          string
+	LineNum       uint
+	Module        string
+	ModuleVersion string
+}
 
-// Returns the name of the function, including module if applicable, that called GetCallerFunction.
-// skipFrames determines how many frames above that to skip.  If you want to know who is calling your
+var moduleLookupByPC = make(map[uintptr]CallerInfo)
+var moduleLookupLock = sync.RWMutex{}
+
+var buildInfo *debug.BuildInfo
+
+//nolint:gochecknoinits // Why: Because we need it and the side effects are self-contained and non-mutating
+func init() {
+	if bi, ok := debug.ReadBuildInfo(); ok {
+		buildInfo = bi
+	}
+}
+
+// GetCallerInfo returns various types of info about the caller of the function.
+// skipFrames determines how many frames above the caller to skip.  If you want to know who is calling your
 // function, you would pass in skipFrames of 1.  skipFrames of 0 will return yourself.
-func GetCallerFunction(skipFrames uint16) (string, error) {
+func GetCallerInfo(skipFrames uint16) (CallerInfo, error) {
 	// We only care about the one stack frame above the caller, so skip at least two:
 	// 1. runtime.Callers
 	// 2. GetCallerModule
@@ -24,7 +45,7 @@ func GetCallerFunction(skipFrames uint16) (string, error) {
 	skipTotal := 2 + int(skipFrames)
 	num := runtime.Callers(skipTotal, pc)
 	if num == 0 {
-		return "", fmt.Errorf("no frames returned from skip %d", skipTotal)
+		return CallerInfo{}, fmt.Errorf("no frames returned from skip %d", skipTotal)
 	}
 
 	if mod, valid := moduleLookupByPC[pc[0]]; valid {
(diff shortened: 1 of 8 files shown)


pkg/callerinfo/callerinfo.go
```diff
@@ -0,0 +1,42 @@
...
+		// Found it in the cache
+		return mod, nil
+	}
+
+	// Not cached -- have to do the slow lookup
+	frames := runtime.CallersFrames(pc)
+	frame, _ := frames.Next()
+
+	// Cache for later
+	moduleLookupByPC[pc[0]] = frame.Function
```
jkinkead, line 40: You need a lock around the write here.

pkg/log/log.go
```diff
@@ -150,6 +151,29 @@ func format(msg, level string, ts time.Time, appInfo Marshaler, mm Many) string
...
+	caller, err := callerinfo.GetCallerFunction(2)
+	switch {
+	case err != nil:
+		entry["source"] = "error"
+	case strings.HasPrefix(caller, "github.com/"):
+		// Example response: github.com/getoutreach/gobox/pkg/callerinfo.Test_Callers
+		splits := strings.Split(caller, "/")
+		// We're going for extracting "getoutreach/gobox" in the above example
+		if len(splits) >= 3 {
+			entry["source"] = splits[1] + "/" + splits[2]
```
jkinkead, line 165: I would not strip off `github.com`, as it's part of the module name.

pkg/log/log.go
```diff
@@ -150,6 +151,29 @@ func format(msg, level string, ts time.Time, appInfo Marshaler, mm Many) string
...
+		} else {
+			entry["source"] = splits[1]
+		}
+	case strings.HasPrefix(caller, "main"):
+		// Main entry point for the app shows up as "main.[funcname]", so just use the app name since we unfortunately
+		// don't have a full package name
+		entry["source"] = app.Info().Name
```
jkinkead, line 172: Prefix with `app:` to help disambiguate?

deregtd: i like that, will do

pkg/log/log.go
jaredallard: I really wish we had a way to get the canonical module name from https://pkg.go.dev/runtime/debug#ReadBuildInfo...

pkg/log/log.go
jaredallard: Or should `source` be the module name as it appears to Go? That goes back to my earlier comment about https://pkg.go.dev/runtime/debug#ReadBuildInfo.

pkg/callerinfo/callerinfo.go
jaredallard: Should we consider having this be an LRU cache?

pkg/log/log.go
jaredallard: Noted this in a DM as well, but we could use the aforementioned build information to retrieve the version of the library in use and store it in a separate field like `source_version`.

pkg/callerinfo/callerinfo.go
deregtd: @jkinkead oh goddamnit, I thought it was both read and write concurrency safe, but it's only read.  Sigh.  Thanks.

@jaredallard It's capped to a count of all of the locations in the app making a log send, so I didn't feel like it made sense to add all the overhead of an LRU cache here.  Even if you have an outstandingly loud app with thousands of log locations, it's still going to be, worst case, like 100-200k of memory.

jaredallard: Makes sense, just wanted to double-check that. I figured I was trying to overoptimize here :D

pkg/log/log.go
```diff
@@ -150,6 +151,29 @@ func format(msg, level string, ts time.Time, appInfo Marshaler, mm Many) string
 		generateFatalFields(entry)
 	}
 
+	// Attempt to map the caller of the log function into the "source" field for identifying if a service or a module
```
jaredallard, line 154: nit: I'd suggest pulling this out into its own function so we reduce the bloat of this function. That being said, I don't really care about this logger at all and want to kill it with fire as soon as possible so I am happy to ignore this

deregtd: ya totes fair

pkg/log/log.go
deregtd: done -- reworked the whole thing

jkinkead (approved):
LGTM although we might be building in some custom labeling we'll need to retrofit into another solution like logrus.

pkg/callerinfo/callerinfo.go
```diff
@@ -0,0 +1,108 @@
...
+package callerinfo
+
+import (
+	"fmt"
+	"runtime"
+	"runtime/debug"
+	"strings"
+	"sync"
+)
+
```
jaredallard, line 16: exported, added comment :D

deregtd: oh, rip good clal

pkg/callerinfo/callerinfo.go
```diff
@@ -0,0 +1,108 @@
...
+	frame, _ := frames.Next()
+
+	ci := CallerInfo{
+		Function: frame.Function,
+		File:     frame.File,
+		LineNum:  uint(frame.Line),
+	}
+
+	// Attempt to back-calc module name by searching through deps
+	if buildInfo != nil {
```
jaredallard, line 67: def move this into another function, having it be nested makes it harder to read here to begin w/ but also it fits into it's own function p easily

pkg/callerinfo/callerinfo.go
```diff
@@ -0,0 +1,108 @@
...
+		for _, mod := range buildInfo.Deps {
+			if strings.HasPrefix(ci.Function, mod.Path) {
+				ci.Module = mod.Path
+				ci.ModuleVersion = mod.Version
+				break
+			}
+		}
+
+		// If we can't find it in the dep list, it must be from the main app
+		if ci.Module == "" {
```
jaredallard, line 77: This seems like an unsafe assumption, maybe we could at least see if `Main.Path` is in the string?

deregtd: It's been true in all my circumstances.  the function shows up as, for example, "main.main", so there's no path prefix :(

jaredallard: Lame, ok. I guess we gotta :(

Can fix it later p easily I guess

jaredallard (approved):

pkg/callerinfo/callerinfo.go
deregtd: yeah, we'll see how it looks when we go wider.  this is all heuristicy :/

pkg/callerinfo/callerinfo.go
deregtd: okie done

PR #268, opened 2023-03-02
feat(logfile): add automatic logfile recorder/player for CLIs

jaredallard:

This PR adds an automatic cli logfile generator as well as player for said log files

I also removed the current tracing and logging code in the current CLI helpers, as well as better organized it. The logfile work is contained to `pkg/cli/logfile`

This takes over from https://github.com/getoutreach/gobox/pull/236


jaredallard:
Used a `replace` directive on `stencil` and this worked OOTB! :tada:

I'm going to try to get this into some weird scenarios, but after that I think we can honestly just merge this as-is. I'll work on finishing the daemon component and getting that out there.

jaredallard:
If you have any comments @hackeryarn 😛

pkg/trace/trace.go
```diff
@@ -111,19 +112,19 @@ var defaultTracer tracer
 //
 // This should be called at the start of the application.
 func StartTracing(serviceName string) error {
-	if err := setDefaultTracer(); err != nil {
+	if err := setDefaultTracer(serviceName); err != nil {
```
hackeryarn, line 115: Once this is released, it would be nice to test this on a couple services just to validate it's 100% correct. This is a slight modification, `StartTracing` is used everywhere.

hackeryarn (approved):
Nope, everything looks good, as I remember it :smile:

tonymbugua-outreach (approved):

jaredallard:
FYI: I built Discovery with this and [traces still work](https://ui.honeycomb.io/outreach-a0/datasets/dev/result/8iUx4HYaprJ/trace/n4RwHY5xYaE?fields[]=c_name&fields[]=c_service_name&span=a3bf54887e1d36cb). So, that makes me confident this won't impact internal services.

PR #270, opened 2023-03-02
fix(shuffler): deprecate shuffler package, allow build when or_e2e is set

jaredallard:

Deprecates the `shuffler` package in favor of native go testing (shuffling is supported OOTB since 1.17 and we've enabled it). Also removes `!or_e2e` constraint due to this causing compilation issues in some packages.

## Jira ID

[XX-XX]

## Notes for your reviewers

[**Internal Slack Thread**](https://outreach-hq.slack.com/archives/C03RV4VGAE7/p1677791110515529)


kaldorn (approved):

tonymbugua-outreach (approved):
LGTM - deprecates shuffler

PR #271, opened 2023-03-07
fix(tracing): Correct YAML value for otel collector tracing

kaldorn:
## Jira ID

[XX-XX]


tonymbugua-outreach (approved):
LGTM - correct yaml

jaredallard (approved):
Pour one out

PR #273, opened 2023-03-07
fix(otel-tracing): use insecure

kaldorn:
We're not sending any traffic outside the cluster to the otel collector, and we don't have TLS termination confgured for this receiver. 

## Jira ID

[XX-XX]


tonymbugua-outreach (approved):
LGTM - uses insecure

jaredallard (approved):
Makes sense with our standards on NOT using TLS within our internal networks.

PR #275, opened 2023-03-13
feat: bad request error

pavelsmejkal:

This PR are aiming to add a generic bad request validation that complies with localization requirements and propagates from the grpc server to its client. 

https://github.com/getoutreach/gobox/pull/275

https://github.com/getoutreach/services/pull/395

## Jira ID

https://outreach-io.atlassian.net/browse/ES-3183

https://github.com/getoutreach/services/pull/395/files

### DIFF
--- a/pkg/orerr/bad_request.go
+++ b/pkg/orerr/bad_request.go
@@ -26,26 +26,26 @@ type Violation struct {
 }
 
 // NewViolation creates a new intance of Violation
-func NewViolation(reason string) *Violation {
-	return &Violation{
+func NewViolation(reason string) Violation {
+	return Violation{
 		Reason: reason,
 	}
 }
 
 // WithField allows to specify field path of the field that violation belongs to
-func (v *Violation) WithField(field string) *Violation {
+func (v Violation) WithField(field string) Violation {
 	v.Field = &field
 	return v
 }
 
 // WithField allows to specify domain name of the field that violation belongs to
-func (v *Violation) WithDomain(domain string) *Violation {
+func (v Violation) WithDomain(domain string) Violation {
 	v.Domain = &domain
 	return v
 }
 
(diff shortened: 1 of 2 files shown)


pkg/orerr/bad_request.go
```diff
@@ -0,0 +1,86 @@
...
+	err = New(err, WithStatus(statuscodes.BadRequest))
+	return &BadRequestError{
+		Err:        err,
+		Violations: violations,
+	}
+}
+
+// Error implements the err interface.
+func (e BadRequestError) Error() string {
+	return fmt.Sprintf("bad request: %s", e.Err.Error())
```
jaredallard, line 74: nit: small preference

```suggestion
	return errors.Wrap(err, "bad request")
```

pavelsmejkal: This function return a string. Do you want me to do it this way?
errors.Wrap(err, "bad request").Error()

jaredallard: Ah, yeah, fair point. Doing `.Error()` is meh... so, we can just ignore this then.

jaredallard (approved):

adrianandrascik-outreach (approved):

pkg/orerr/bad_request.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Reason of the error. In most of the cases the validation rule indentifier.
+	Reason string
+
+	// Additional structured details about this error. Could be used for localization of the error.
+	Metadata map[string]string
+}
+
+// NewViolation creates a new intance of Violation
+func NewViolation(reason string) *Violation {
```
adrianandrascik-outreach, line 29: consider using value semantics, since a slice of pointers is quite cumbersome to work with later on

PR #276, opened 2023-03-16
fix(trace): usage revealed a panic

AndrewWinterman:

passing events.NewErrorInfo(nil), a totally valid use case, was sufficient to cause a panic. This adds a smatterring of nil checks in the type switch to ensure we don't hit that.

## Jira ID

[XX-XX]

### DIFF
--- a/pkg/trace/otel.go
+++ b/pkg/trace/otel.go
@@ -226,9 +226,6 @@ func (t *otelTracer) end(ctx context.Context) {
 func (t *otelTracer) addInfo(ctx context.Context, args ...log.Marshaler) {
 	if span := trace.SpanFromContext(ctx); span != nil {
 		for _, arg := range args {
-			if arg == nil {
-				continue
-			}
 			kvs := marshalToKeyValue(arg)
 			span.SetAttributes(kvs...)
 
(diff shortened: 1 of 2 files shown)


pkg/trace/otel.go
```diff
@@ -226,16 +226,27 @@ func (t *otelTracer) end(ctx context.Context) {
...
+			if arg == nil {
+				continue
+			}
 			kvs := marshalToKeyValue(arg)
 			span.SetAttributes(kvs...)
 
 			switch v := arg.(type) {
 			case *events.ErrorInfo:
+				if v == nil {
+					//
```
asms, line 238:
```suggestion
```

pkg/trace/otel.go
```diff
@@ -226,16 +226,27 @@ func (t *otelTracer) end(ctx context.Context) {
 func (t *otelTracer) addInfo(ctx context.Context, args ...log.Marshaler) {
 	if span := trace.SpanFromContext(ctx); span != nil {
 		for _, arg := range args {
+			if arg == nil {
```
asms, line 229: I don't think `arg` will ever be `nil`.
https://go.dev/play/p/r7lBirpAWCM

asms (approved):
LGTM. Minor comments.

pkg/trace/otel.go
AndrewWinterman: thanks!

chatchai-outreach (approved):

PR #288, opened 2023-03-23
fix: trace logFile

tonymbugua-outreach:
fix issue where clients are broken sometimes ([thread](https://outreach-hq.slack.com/archives/CN9MU7GLW/p1679596508166189))

## Jira ID

[XX-XX]


jkinkead (approved):

kaldorn (approved):
One approval is good, but two is better.

PR #300, opened 2023-04-13
feat(trace): Support disabling logging for trace call methods

jaredallard:

This PR adds a `CallOptions` type that can be used for specific options for tracing calls. Ideally, this would be limited to artifacts of traces (e.g., logs). The first, and only, option implemented is `DisableInfoLogging`. This defaults to `false`. When set to `true`, info logs are not generated when a call ends.'

I also cleaned up some tests to no longer be attached to the test suite struct invoked by `shuffler.Run()`. This helps us when we fully remove the shuffler in favor of using standard go testing.

Updated the `gopls` build flags to not include `or_e2e` for invocation of tests through VSCode and compilation. This doesn't really matter except for some packages, ideally we'll remove these tags entirely in the future when we stop using build tags to handle configuration loading logic (e.g., move entirely to devspace).

## Jira ID

[DT-3673]

[DT-3673]: https://outreach-io.atlassian.net/browse/DT-3673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -31,7 +31,7 @@
     "editor.defaultFormatter": "redhat.vscode-yaml"
   },
   "gopls": {
-    "build.buildFlags": ["-tags=or_test,or_dev,or_e2e,or_int"]
+    "build.buildFlags": ["-tags=or_test,or_dev,or_int"]
   },
   "[terraform]": {
     "editor.defaultFormatter": "hashicorp.terraform"
--- a/internal/call/call.go
+++ b/internal/call/call.go
@@ -16,9 +16,10 @@ import (
 	"github.com/getoutreach/gobox/pkg/metrics"
 )
 
-// Type tracks the call type.
+// Type tracks the type of calling being made.
 type Type string
 
+// Contains the call type constants.
 const (
 	// TypeHTTP is a constant that denotes the call type being an HTTP
 	// request.
@@ -35,15 +36,29 @@ const (
 
 // Info tracks information about an ongoing synchronous call.
 type Info struct {
+	// Name is the name of the call.
 	Name string
+
+	// Type is the type of the call.
 	Type Type
+
+	// Opts are the options for this call.
+	Opts Options
+
+	// Kind is the type of call being made. See metrics.CallKind for more
+	// information.
 	Kind metrics.CallKind
+
+	// Args are the arguments to the call, this is used for attributes on
+	// logs and traces.
 	Args []logf.Marshaler
+
+	// ErrInfo is the information for an error that occurred during the call.
+	// This is set by SetStatus and used for reporting that an error occurred.
+	ErrInfo *events.ErrorInfo
 
 	events.Times
 	events.Durations
-
-	ErrInfo *events.ErrorInfo
 
 	mu sync.Mutex
 }
@@ -117,8 +132,7 @@ func (info *Info) MarshalLog(addField func(key string, value interface{})) {
 }
 
 // Tracker helps manage a call info via the context.
-type Tracker struct {
-}
+type Tracker struct{}
 
 // StartCall creates a new call Info object and returns a new context
 // where tracker.Info(ctx) will return the newly setup call Info object.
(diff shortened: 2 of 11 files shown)


internal/call/call.go
```diff
@@ -117,8 +117,7 @@ func (info *Info) MarshalLog(addField func(key string, value interface{})) {
 }
 
 // Tracker helps manage a call info via the context.
-type Tracker struct {
-}
+type Tracker struct{}
```
george-e-shaw-iv, line 120: This is a good idea to consolidate this onto one line. Nice find.

.vscode/settings.json
```diff
@@ -31,7 +31,7 @@
     "editor.defaultFormatter": "redhat.vscode-yaml"
   },
   "gopls": {
-    "build.buildFlags": ["-tags=or_test,or_dev,or_e2e,or_int"]
+    "build.buildFlags": ["-tags=or_test,or_dev,or_int"]
```
jkinkead, line 34: What is this doing? It's not in the PR description. Add it?

internal/call/option.go
```diff
@@ -5,5 +5,13 @@ package call
 // Option defines the call Info adjustment function
 type Option func(c *Info)
 
+// Options contains options for all tracing calls.
+type Options struct {
+	// DisableInfoLogging determines if info logging should be disabled or not
+	// when a call is finished. This is useful for calls that are expected to
+	// be very frequent, such as HTTP requests.
```
jkinkead, line 12: Be explicit with booleans:
```suggestion
	// DisableInfoLogging turns off per-call info logging if set to true. If false, every successful
	// (statuscodes.CategoryOK) call will have an Info line emitted.
```

pkg/trace/call.go
```diff
@@ -107,6 +106,13 @@ func SetCallError(ctx context.Context, err error) error {
 // rethrows them.  Any panics are converted to errors and cause error
 // logging to happen (as do any SetCallStatus calls)
 func EndCall(ctx context.Context) {
+	callInfo := callTracker.Info(ctx)
+	if callInfo == nil {
+		// There was no call associated with this context, so we can't
+		// do anything.
```
jkinkead, line 112: Is it correct to skip the `defer End(ctx)` in this case? I assume it's a no-op without a call?

pkg/trace/call_option_test.go
```diff
@@ -38,3 +44,61 @@ func TestAsOutboundCall(t *testing.T) {
...
+	// Test that the default is false
+	callInfo := startCall(func(c *call.Info) {})
+	assert.Equal(t, false, callInfo.Opts.DisableInfoLogging)
+
+	// Test that trace.WithInfoLoggingDisabled() sets the DisableInfoLogging
+	// option to true.
+	callInfo = startCall(trace.WithInfoLoggingDisabled())
+	assert.Equal(t, true, callInfo.Opts.DisableInfoLogging)
+
+	// Make a call and ensure that info logs are not emitted.
```
jkinkead, line 67: Is there a test that logs _are_ emitted by default?

pkg/trace/call_test.go
```diff
@@ -37,7 +37,7 @@ func (m *Model) MarshalLog(addField func(k string, v interface{})) {
 	addField("model.id", m.ID)
 }
 
-func (suite) TestNestedCall(t *testing.T) {
+func TestNestedCall(t *testing.T) {
```
jkinkead, line 40: Add a short note in the PR description about this cleanup?

jkinkead (approved):
LGTM! Just some requests for docs cleanup, and one question / suggestion for another test case.

pkg/trace/call.go
jaredallard: Yes, while it would noop it doesn't make sense to hit that code path if we have no trace information -- that's equivalent to having no trace at all, so one can't end it.

.vscode/settings.json
jaredallard: Done

tonymbugua-outreach (approved):

PR #301, opened 2023-04-18
fix(logfile): handle []interface{} stringslice attribute conversion

jaredallard:

JSON will sometimes convert mixed `stringslice` attributes into `[]interface{}` instead of `[]string`. This updates our attribute transformation logic to handle this, and to be more robust in general with tests and some other edgecase handling (e.g., `int64` always being `float64` might not be the case forever.)

## Jira ID

[XX-XX]


tonymbugua-outreach (approved):
LGTM - fix panic bug

PR #308, opened 2023-05-18
fix(delibird): pass stdin to subproccess when not term

jaredallard:

This PR fixes the stdin logic to properly pass stdin to the created sub-process when running not as a terminal. When running in a terminal, this is properly handled by the `ptyOutputHook` function.

## Jira ID

[DT-3742]

[DT-3742]: https://outreach-io.atlassian.net/browse/DT-3742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/cli/logfile/logfile_unix.go
+++ b/pkg/cli/logfile/logfile_unix.go
@@ -190,16 +190,16 @@ func ptyOutputHook(l net.Listener, cmd *exec.Cmd, ptmx,
 		return nil, errors.Wrap(err, "failed to attach stdin to pty")
 	}
 
-	// forward os.Stdin to the PTY
-	//nolint:errcheck // Why: Best effort
-	go io.Copy(ptmx, os.Stdin)
-
 	finishedChan := make(chan struct{})
 
 	w, h, err := term.GetSize(int(os.Stdin.Fd()))
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get terminal size")
 	}
+
+	// forward os.Stdin to the PTY
+	//nolint:errcheck // Why: Best effort
+	go io.Copy(ptmx, os.Stdin)
 
 	rec := newRecorder(logFile, w, h, cmd.Path, cmd.Args[1:], l)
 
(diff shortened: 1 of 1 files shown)


pkg/cli/logfile/logfile_unix.go
```diff
@@ -101,6 +101,7 @@ func Hook() error {
 
 		cmd.Stdout = io.MultiWriter(os.Stdout, rec)
 		cmd.Stderr = io.MultiWriter(os.Stderr, rec)
+		cmd.Stdin = os.Stdin
```
jkinkead, line 104: The error was being triggered on line 84, so this isn't fixing the bug you think it is.

jkinkead (approved):
Changes look good, but this isn't fixing the root cause of the terminal width not being discoverable.

pkg/cli/logfile/logfile_unix.go
```diff
@@ -33,7 +33,8 @@ func Hook() error {
 		return nil
 	}
 
-	isTerminal := term.IsTerminal(int(os.Stdout.Fd()))
+	// if both stdin and stdout are not terminals, then we don't need a pty
```
jkinkead, line 36: Why do we need a terminal _at all_? I can't remember this.

jkinkead (commented):
Is there any way to plug in a simple end-to-end test here? Not an Outreach "e2e", but like:

1. Build a simple binary that uses this library.
2. Verify that you can pipe stuff to it / redirect from it.

pkg/cli/logfile/logfile_unix.go
jaredallard: If we pipe to a file, terminal checks will fail. This means that logs will never be formatted. We use a pty to ensure that users can still use things that need user input and terminal drawing (e.g., `devenv top`) and have nicely formatted logs.

jaredallard:
> Is there any way to plug in a simple end-to-end test here? Not an Outreach "e2e", but like:
> 
> 1. Build a simple binary that uses this library.
> 2. Verify that you can pipe stuff to it / redirect from it.

@jkinkead  Yes, we could add it. I'd prefer to do it in another ticket though given the rest of this functionality has been tested already and this fixes a small bug in the original implementation. (I'll pick it up soon but don't want to block all of this on that, if possible.

jkinkead (approved):
LGTM, thanks for the answers!

tonymbugua-outreach (approved):

PR #309, opened 2023-05-19
fix(trace): only skip emitting emitting logs/metrics when callinfo is not set

jaredallard:

Currently we're having an issue with metrics "merging", this is a shot at solving this problem by refining the original intention of short-circuting the `callInfo` logic.

Instead of returning immediately, we opt to not schedule the `defer` function that emits metrics/logs on call information. Likely what was happening is that any traces without
call information were not being ended and thus "merging" ocurred.

## Jira ID

[DT-3752]

## Notes for your reviewers

[DT-3752]: https://outreach-io.atlassian.net/browse/DT-3752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/trace/call.go
+++ b/pkg/trace/call.go
@@ -107,33 +107,30 @@ func SetCallError(ctx context.Context, err error) error {
 // logging to happen (as do any SetCallStatus calls)
 func EndCall(ctx context.Context) {
 	callInfo := callTracker.Info(ctx)
-	if callInfo == nil {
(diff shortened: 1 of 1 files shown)


tonymbugua-outreach (approved):

jaredallard:
Didn't work.

PR #311, opened 2023-05-19
fix(trace): restore previous callinfo check behaviour

jaredallard:

Another attempt at fixing [DT-3752], this is the only real change I can tell that happened that could potentially break anything. `addDefaultTracerInfo` seems like it could be connected, but I cannot for the life of me figure out what that code is doing... so here's my attempt at letting it be and maybe that'll somehow fix the bug.

## Jira ID

[DT-3752]

## Notes for your reviewers

[DT-3752]: https://outreach-io.atlassian.net/browse/DT-3752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DT-3752]: https://outreach-io.atlassian.net/browse/DT-3752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/trace/call.go
+++ b/pkg/trace/call.go
@@ -106,13 +106,6 @@ func SetCallError(ctx context.Context, err error) error {
 // rethrows them.  Any panics are converted to errors and cause error
 // logging to happen (as do any SetCallStatus calls)
 func EndCall(ctx context.Context) {
-	callInfo := callTracker.Info(ctx)
-	if callInfo == nil {
-		// There was no call associated with this context, so we can't
-		// do anything.
-		return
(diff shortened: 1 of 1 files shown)


jaredallard:
Haven't been able to test this build yet.

jaredallard:
Nope.

PR #312, opened 2023-05-23
feat: support disabling logging for trace call methods

jaredallard:
This PR adds back the functionality added in #300. This part of the code was never broken, so nothing needs to be modified.

[DT-3758]

[DT-3758]: https://outreach-io.atlassian.net/browse/DT-3758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/trace/call.go
+++ b/pkg/trace/call.go
@@ -107,35 +107,33 @@ func SetCallError(ctx context.Context, err error) error {
 // logging to happen (as do any SetCallStatus calls)
 func EndCall(ctx context.Context) {
 	callInfo := callTracker.Info(ctx)
-	if callInfo == nil {
-		// There was no call associated with this context, so we can't
-		// do anything.
-		return
-	}
-
 	defer End(ctx)
 
-	defer func(info *call.Info) {
-		addDefaultTracerInfo(ctx, info)
-		info.ReportLatency(ctx)
+	// If we don't have any call information we don't need to
+	// emit any logs, metrics, or traces.
+	if callInfo != nil {
+		defer func(info *call.Info) {
+			addDefaultTracerInfo(ctx, info)
+			info.ReportLatency(ctx)
 
-		if info.ErrInfo != nil {
-			switch category := orerr.ExtractErrorStatusCategory(info.ErrInfo.RawError); category {
-			case statuscodes.CategoryClientError:
-				log.Warn(ctx, info.Name, info, IDs(ctx), traceEventMarker{})
-			case statuscodes.CategoryServerError:
-				log.Error(ctx, info.Name, info, IDs(ctx), traceEventMarker{})
-			case statuscodes.CategoryOK: // just in case if someone will return non-nil error on success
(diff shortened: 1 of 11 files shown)


pkg/trace/call.go
```diff
@@ -107,6 +106,13 @@ func SetCallError(ctx context.Context, err error) error {
...
 // logging to happen (as do any SetCallStatus calls)
 func EndCall(ctx context.Context) {
+	callInfo := callTracker.Info(ctx)
+	if callInfo == nil {
+		// There was no call associated with this context, so we can't
+		// do anything.
+		return
+	}
+
 	defer End(ctx)
```
jkinkead, line 116: I think you should still have this deferred even with a nil `callInfo`, to try to maintain compatibility as much as possible. This function doesn't rely on the info, only the span, so it wouldn't have panicked before.

pkg/trace/call_option_test.go
```diff
@@ -38,3 +44,61 @@ func TestAsOutboundCall(t *testing.T) {
...
+	trace.EndCall(ctx)
+
+	if diff := cmp.Diff([]logf.F(nil), recorder.Entries(), differs.Custom()); diff != "" {
+		t.Fatal("unexpected events", diff)
+	}
+
+	// now make a call with info logging enabled and ensure that info logs are
+	// emitted.
+	ctx = trace.StartCall(context.Background(), "test")
+	trace.EndCall(ctx)
```
jkinkead, line 81: Can you throw in a test for your new nil handling code as well?

jkinkead (commented):
A couple small comments.

pkg/trace/call.go
jaredallard: If `callInfo` is `nil`, this does panic actually. `End()` ends up using it later in the call stack. I don't know of an easy way to avoid that yet, but just noting that it is somewhat dangerous right now.

jaredallard: I was wrong, `callTracker.EndCall` does, not `End` itself.

pkg/trace/call_option.go
```diff
@@ -5,35 +5,70 @@
...
+// AsHTTPCall changes the call type to HTTP. This SHOULD NOT be used for outbound
+// calls to external services. Use AsOutboundCall instead.
 func AsHTTPCall() call.Option {
 	return func(c *call.Info) {
 		c.Type = call.TypeHTTP
 	}
 }
 
-// AsOutboundCall set the call type to Outbound
+// AsOutboundCall set the call type to Outbound. This is meant for calls
```
jkinkead, line 46: hobgoblining:
```suggestion
// AsOutboundCall changes the call type to Outbound. This is meant for calls
```

jkinkead (approved):
LGTM, one consistency nit.

PR #315, opened 2023-05-25
feat: add status codes for deadline and canceled

milankonir:

I'm migrating `mailroomapi` service from returning grpc status codes directly to gobox error codes in order to make the tracing properly (even client side errors are logged as server errors into telemetry now).

To support all of the current grpc codes used by `mailroomapi`, I'm adding two new codes for `DeadlineExceeded` and `Canceled`. 

## Jira ID

[XX-XX]

### DIFF
--- a/pkg/statuscodes/statuscodes.go
+++ b/pkg/statuscodes/statuscodes.go
@@ -15,12 +15,14 @@ const (
 	OK StatusCode = 600
 
 	// Client-caused error responses
-	BadRequest   StatusCode = 700
-	Unauthorized StatusCode = 701
-	Forbidden    StatusCode = 702
-	NotFound     StatusCode = 703
-	Conflict     StatusCode = 704
-	RateLimited  StatusCode = 705
+	BadRequest       StatusCode = 700
+	Unauthorized     StatusCode = 701
+	Forbidden        StatusCode = 702
+	NotFound         StatusCode = 703
+	Conflict         StatusCode = 704
+	RateLimited      StatusCode = 705
+	DeadlineExceeded StatusCode = 706
+	Canceled         StatusCode = 707
 
 	// Server-caused error responses
 	InternalServerError StatusCode = 800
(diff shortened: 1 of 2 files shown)


pkg/statuscodes/statuscodes.go
```diff
@@ -15,12 +15,14 @@ const (
...
-	NotFound     StatusCode = 703
-	Conflict     StatusCode = 704
-	RateLimited  StatusCode = 705
+	BadRequest       StatusCode = 700
+	Unauthorized     StatusCode = 701
+	Forbidden        StatusCode = 702
+	NotFound         StatusCode = 703
+	Conflict         StatusCode = 704
+	RateLimited      StatusCode = 705
+	DeadlineExceeded StatusCode = 706
```
deregtd, line 24: These are golang notions, not networking notions.  I can see adding a "TimedOut" response, but "canceled" doesn't mean anything to the caller.  Why was it canceled, and what was the effect on my request?  An internal error that caused a canceled can return InternalServerError.  But this protocol is intentionally obfuscating golang errors -- this is experience-based, not language-based.

deregtd (changes requested):

pkg/statuscodes/statuscodes.go
adrianandrascik-outreach: i agree that naming could be made more generic at least for the `DeadlineExceeded`, but these are not golang notions, but gRPC API notions AFAIK as you can see here https://grpc.github.io/grpc/core/md_doc_statuscodes.html

deregtd: Yeah -- we're just trying to make a more intent-based subset of response codes.  Both HTTP and GRPC have a ton of detailed codes that mostly aren't providing value to callers, especially for internal S2S calls.  I don't believe we have a single client-cancelable action in our entire service codebase yet.  So I don't know what we would need a ClientCanceled status code for, and if we add it now, people are going to add it as a response to map from "context canceled", which is actually an internal bug/issue local to the service, which the caller doesn't care about.

We're trying to intentionally limit the response code surface area to force service developers to bucketize issues into the very few buckets that callers actually care about differentiating.

PR #318, opened 2023-06-14
feat: allow self-classification of errors

alexandrkara-outreach:

Allow errors to implement:

```golang
StatusCategory() statuscodes.StatusCategory
```

and if it's implemented, use that category. This allows for greater flexibility when deciding how to categorize errors that don't fit the predefined status codes.

## Jira ID

[XX-XX]

## Notes for your reviewers

TESTED=go test -v -race -count=1 -tags=or_test,or_int ./pkg/orerr/...

### DIFF
--- a/pkg/orerr/statuscodewrap.go
+++ b/pkg/orerr/statuscodewrap.go
@@ -10,6 +10,10 @@ import (
 	"github.com/getoutreach/gobox/pkg/statuscodes"
 )
 
+type StatusCodeProvider interface {
+	StatusCode() statuscodes.StatusCode
+}
+
 type StatusCategoryProvider interface {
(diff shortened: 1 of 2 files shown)


deregtd:
I'm confused about the intent here -- I'm okay allowing people to make their own providers if they'd rather do that than wrap the statusprovider object we provide, but you should be required to expose both a code and a category at the same time.  This is going to result in some really weird errors that have a category but no code, and that might even make the logging library crash, I'm not sure.

jaredallard:
@alexandrkara-outreach Do you plan on continuing this work? Or should we close it for now? Thank you!! 🙇

jaredallard:
Going to close for now. Please feel free to reopen it if necessary. Thank you!

PR #319, opened 2023-06-27
feat: Add Unprocessable entity

romain-tadiello:

This kind of error code can be returned by APIv2 though we don't have any mapping for it. 

## Jira ID

[XX-XX]


ChrisPRobinson (approved):
LGTM

dundee (approved):
LGTM

zaquestion (approved):

deregtd:
The point of these error codes in gobox is to have a very high level code to bucketize for instrumentation. Something specific like "unprocessable entity" is a service-specific error code and should be returned as a sub-error.

PR #322, opened 2023-07-05
feat: Updating otel to latest versions, fixing unit tests

deregtd:
To fix some dependency issues in vault-client, I had to walk back up the chain to gobox to update to the latest otel libraries.  This caused several changes in attributes that the library is sending out, changing the unit test expected results.  It's unclear if this is idiomatic changes or what, so I'm not sure how to review this.


richardlarocque (approved):
I did some digging.  There have been some changes in the naming conventions, and they do look related to these diffs.

Here's a changelog: https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md.

I also stumbled across this feature, which may or may not be available in whatever version of Golang otel we're using.  I'm guessing it's not worth the effort to try to maintain compatibility here, though.  Our Node.js, Rust, and Ruby implementations all look different anyway.  This won't make the existing situation much worse.

https://github.com/open-telemetry/opentelemetry-specification/pull/3443/files

yj-yan (approved):

PR #325, opened 2023-07-18
feat: port context-aware cond to gobox

AndrewWinterman:
I've been using this for a few months in polaroid (non-generic version), so I figured it's reasonable at this point to port it to GoBox,

## Jira ID

[XX-XX]

### DIFF
--- a/pkg/async/cond.go
+++ b/pkg/async/cond.go
@@ -1,0 +1,94 @@
+// Copyright 2023 Outreach Corporation. All Rights Reserved.
+
+// Description: Cond.go provides a context respecting sync condition
+
+package async
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+)
+
+// Cond is a sync.Cond that respects context cancellation.
+// It provides equivalent functionality to sync.Cond (excepting there is no `Signal` method), except that
+// the Wait method exits with error if the context cancels.
+type Cond struct {
+	pointer atomic.Pointer[chan struct{}]
+	mu      sync.Mutex
+}
+
+// ch returns the channel that Waiters are waiting on, possibly creating one if it doesn't exist.
+func (c *Cond) ch() chan struct{} {
+	t := make(chan struct{})
+	c.pointer.CompareAndSwap(nil, &t)
+	return *c.pointer.Load()
+}
+
+// Wait waits for the state change Broadcast until context ends.
+func (c *Cond) Wait(ctx context.Context) error {
+	select {
+	case <-c.ch():
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+// Broadcast signals the state change to all Waiters
+func (c *Cond) Broadcast() {
+	ch := c.ch()
+	// now that we retrieved the channel, new calls to Wait should get a new channel
+	c.pointer.Store(nil)
+	close(ch)
+}
+
+// WaitForCondition checks if the condition is true or the context is done, otherwise
+// it waits for the state change Broadcast.
+//
+// if it returns without error, it also locks the provided locker and the caller must call the returned function
+// to unlock it. Until they call unlock, the state should not be changed.
+func (c *Cond) WaitForCondition(ctx context.Context, locker sync.Locker, condition func() bool) (func(), error) {
+	for {
+		err := c.lockWithContext(ctx, locker)
+		if err != nil {
+			return func() {}, err
+		}
+
+		// we have the lock, we can safely check the condition
+		ok := condition()
+
+		if !ok {
+			// condition not met
+			// but we acquired the lock. so unlock it...
+			locker.Unlock()
+
+			// either way, wait for the next signal
+			waitErr := c.Wait(ctx)
+			if waitErr != nil {
+				return func() {}, waitErr
+			}
+		} else {
+			// condition met, return the unlock function and nil error
+			// client must call the unlock function to unlock the mutex
+			// client guaranteed the condition holds while mutex lock is held.
+			return locker.Unlock, nil
+		}
+	}
+}
+
+// lockWithContext waits to either acquire the lock or for the context to end.
+// It returns an error if context ends before it can acquire the lock
+func (c *Cond) lockWithContext(ctx context.Context, locker sync.Locker) error {
+	lockAcquired := make(chan struct{})
+	go func() {
+		locker.Lock()
+		close(lockAcquired)
+	}()
+	select {
+	case <-lockAcquired:
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+	return nil
+}
--- a/pkg/async/cond_test.go
+++ b/pkg/async/cond_test.go
@@ -1,0 +1,194 @@
+package async
+
+import (
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/sync/errgroup"
+)
+
+func TestCond(t *testing.T) {
+
+	t.Run("broadcast wakes up waiter", func(t *testing.T) {
+		cond := Cond{}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+		defer cancel()
+
+		go func() {
+			time.Sleep(50 * time.Millisecond) // just a breath so the other goroutine goes first
+			cond.Broadcast()
+		}()
+
+		err := cond.Wait(ctx)
+		assert.Nil(t, err)
+	})
+
+	t.Run("can safely interleave broadcasts", func(t *testing.T) {
+		cond := Cond{}
+		ctx, cancel := context.WithTimeout(context.Background(), 300*time.Millisecond)
+		defer cancel()
+		for j := 0; j < 10; j++ {
+			start := make(chan struct{})
+			g, ctx := errgroup.WithContext(ctx)
+			g.Go(func() error {
+				return cond.Wait(ctx)
+			})
+			for i := 0; i < 10; i++ {
+				g.Go(func() error {
+					<-start
+					cond.Broadcast()
+					return nil
+				})
+			}
+			g.Go(func() error {
+				time.Sleep(10 * time.Millisecond) // just a breath so the other goroutine goes first
+				close(start)
+				return nil
+			})
+			err := g.Wait()
+			assert.Nil(t, err)
+		}
+	})
+
+	t.Run("broadcast wakes all waiters", func(t *testing.T) {
+		cond := Cond{}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+		defer cancel()
+		g, ctx := errgroup.WithContext(ctx)
+		// start everyone waiting
+		for i := 0; i < 10; i++ {
+			g.Go(func() error {
+				return cond.Wait(ctx)
+			})
+		}
+
+		// wake em all up
+		g.Go(func() error {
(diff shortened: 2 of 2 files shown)


pkg/async/cond.go
```diff
@@ -0,0 +1,94 @@
...
+// the Wait method exits with error if the context cancels.
+type Cond struct {
+	pointer atomic.Pointer[chan struct{}]
+}
+
+// ch returns the channel that Waiters are waiting on, possibly creating one if it doesn't exist.
+func (c *Cond) ch() chan struct{} {
+	t := make(chan struct{})
+	c.pointer.CompareAndSwap(nil, &t)
+	return *c.pointer.Load()
```
jkinkead, line 24: This is going to create a lot of garbage objects if `ch()` is invoked in a hot path. Maybe optimize?
```suggestion
	value := c.pointer.Load()
	if value == nil {
		t := make(chan struct{})
		c.pointer.CompareAndSwap(nil, &t)
		value = c.pointer.Load()
	}
	return *value
```

pkg/async/cond.go
```diff
@@ -0,0 +1,94 @@
...
+// Broadcast signals the state change to all Waiters
+func (c *Cond) Broadcast() {
+	// now that we retrieved the channel, new calls to Wait should get a new channel
+	ch := c.pointer.Swap(nil)
+	if ch != nil {
+		close(*ch)
+	}
+}
+
+// WaitForCondition checks if the condition is true or the context is done, otherwise
```
jkinkead, line 46: Document return values?

pkg/async/cond.go
```diff
@@ -0,0 +1,94 @@
...
+// it waits for the state change Broadcast.
+//
+// if it returns without error, it also locks the provided locker and the caller must call the returned function
```
jkinkead, line 49: Clearly document that the lock is held while running `condition`, and released on error?

pkg/async/cond.go
```diff
@@ -0,0 +1,94 @@
...
+func (c *Cond) lockWithContext(ctx context.Context, locker sync.Locker) error {
+	lockAcquired := make(chan struct{})
+	go func() {
+		locker.Lock()
+		close(lockAcquired)
+	}()
+	select {
+	case <-lockAcquired:
+	case <-ctx.Done():
+		return ctx.Err()
```
jkinkead, line 91: Close `lockAcquired` here?

pkg/async/cond.go
```diff
@@ -0,0 +1,94 @@
...
+			return locker.Unlock, nil
+		}
+	}
+}
+
+// lockWithContext waits to either acquire the lock or for the context to end.
+// It returns an error if context ends before it can acquire the lock
+func (c *Cond) lockWithContext(ctx context.Context, locker sync.Locker) error {
+	lockAcquired := make(chan struct{})
+	go func() {
```
jkinkead, line 84: This goroutine is going to leak if the context hits `Done`, particularly if the locker is blocked forever. Can you instead do something like `select` on `locker.Lock()` and reading `lockAcquired`? That way you can close `lockAcquired` below (per my other suggestion) and ensure this goroutine closes out.

jkinkead (commented):
Comments / suggestions.

pkg/async/cond.go
AndrewWinterman: good point, and good idea.

pkg/async/cond.go
AndrewWinterman: I don't think we can lock on `locker.Lock()` unfortunately. It doesn't return anything. I'm not exactly sure how to fix this, because we need whatever is holding the lock to release it when the context ends, or else this goroutine never exits.

I briefly tried to make this work with TryLock instead, which works. Could switch to that

AndrewWinterman:
because Jesse was worried about extra allocations, I aded a couple of benchmarks which also revealed a race that could result in an NPE:

```
BenchmarkCond
BenchmarkCond/one_broadcasts;_one_wait
BenchmarkCond/one_broadcasts;_one_wait-10         	     100	  11171022 ns/op	         1.170 ms_corrected/op	     489 B/op	       9 allocs/op
BenchmarkCond/one_broadcasts;_many_waiters
BenchmarkCond/one_broadcasts;_many_waiters-10     	     100	  11239043 ns/op	         1.230 ms_corrected/op	    2023 B/op	      37 allocs/op
PASS

File: pkg/async/cond.go (hunk 5 above)
Comment (AndrewWinterman) [reply pkg/async/cond.go]: I switched to that.

File: pkg/async/cond.go (hunk 3 above)
Comment (AndrewWinterman) [reply pkg/async/cond.go]: done

File: pkg/async/cond.go (hunk 2 above)
Comment (AndrewWinterman) [reply pkg/async/cond.go]: done

File: pkg/async/cond.go (hunk 1 above)
Comment (AndrewWinterman) [reply pkg/async/cond.go]: i did something a little different, but similar

File: pkg/async/cond.go [hunk 6]
```diff
@@ -0,0 +1,97 @@
...
+	load := c.pointer.Load()
+	if load == nil {
+		t := make(chan struct{})
+		c.pointer.CompareAndSwap(nil, &t)
+		return t
+	}
+	return *load
+}
+
+// Wait waits for the state change Broadcast until context ends.
```
jaredallard, line 37: Note the behaviour of `Wait` when the context cancels? (It returns the context error)

pkg/async/cond.go
```diff
@@ -0,0 +1,97 @@
...
+// or the lock was not available, it releases the locker and waits for the state change Broadcast.
+// If the context ends during any of these operations, the context error is returned.
+//
+// WaitForCondition returns an unlock function that should always be called to unlock the locker.
+// unlock is safe to call regardless  of error.
+// Error should only be returned if the context ends before the condition is met.
+//
+// If it returns without error, it also locks the provided locker and the caller must call the returned function
+// to unlock it. Until they call unlock, the state should not be changed.
+func (c *Cond) WaitForCondition(ctx context.Context, condition func() bool) (unlock func(),
```
jaredallard, line 68: I don't understand the addition of this function. Given the purpose of this package is to "provides a context respecting sync condition", I think we shouldn't deviate from the standard library signature in this package.

We should probably put this in a different package than `async`, maybe `pkg/syncctx`? Or something similar?

jaredallard: To be clear, my issue here is matching the `Cond` signature and adding new functionality. If we want to make this a special package, I'm OK with that, but we should make it clear that our goal is not to provide the stdlib.

AndrewWinterman: I have no problem moving it. 

This function is supposed to encapsulate their instructions on how to use cond, basically you have to invoke wait in a loop, locking when you start the loop, waiting for Wait to return, and then doing your operation before unlocking. That's enough machinery to warrant encapsulation imho, and the lack of that makes it harder to use sync.Cond.

AndrewWinterman: So if you want me to move this function specifically I'm happy to. Maybe just to another file or moving the receiver to a function argument would be sufficient?

jaredallard: I prefer the latter. But also saying that we're not trying to perfectly match `sync.Cond` is also fine! A comment on the `Cond` struct would suffice for me.

AndrewWinterman: I'll comment cond then

AndrewWinterman: done

jaredallard (approved):

pkg/async/cond.go
```diff
@@ -0,0 +1,129 @@
...
+// Cond mimics sync.Cond in purpose, with the added goals of easing usability slightly and respecting context expiry.
+//
+// It provides functionality similar sync.Cond
+//   - there is no signal method for waking a single Waiter
+//   - the Wait method exits with error if the context cancels.
+//   - it provides WaitForCondition, which intends to encapsulate the common pattern of acquiring a lock, checking
+//     a condition, and releasing the lock before waiting for a state change if the condition is not met.
+type Cond struct {
+	pointer atomic.Pointer[chan struct{}]
+	Mu      sync.Mutex
```
jkinkead, line 22: Document public field?

pkg/async/cond.go
```diff
@@ -0,0 +1,129 @@
...
+
+// ch returns the channel that Waiters are waiting on,
+// possibly creating one if it hasn't been initialized
+func (c *Cond) ch() chan struct{} {
+	// non atomic check for nil channel
+	load := c.pointer.Load()
+	if load == nil {
+		t := make(chan struct{})
+		c.pointer.CompareAndSwap(nil, &t)
+		return t
```
jkinkead, line 33: This doesn't handle the race where two threads enter this if block at the same time. You should return the value of `c.pointer.Load()` here (although technically only needed if `CompareAndSwap` returns `false`).

Correct verbose version:
```suggestion
		t := make(chan struct{})
		if c.pointer.CompareAndSwap(nil, &t) {
			return t
		}
		return c.pointer.Load()
```

pkg/async/cond.go
```diff
@@ -0,0 +1,129 @@
...
+	ch := c.ch()
+	select {
+	case <-ch:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+// Broadcast signals the state change to all Waiters
```
jkinkead, line 50: Consistently terminate comment sentences:
```suggestion
// Broadcast signals the state change to all Waiters.
```

pkg/async/cond.go
```diff
@@ -0,0 +1,129 @@
...
+func (c *Cond) Broadcast() {
+	// swap in a new channel, close the old one
+	newChan := make(chan struct{})
+	ch := c.pointer.Swap(&newChan)
+	if ch != nil {
+		close(*ch)
+	}
+}
+
+// WaitForCondition acquires Cond's lock, then checks if the condition is true. If the condition is not true,
```
jkinkead, line 60: Clarify what the lock actually is (and use the receiver name):
```suggestion
// WaitForCondition acquires c.Mu, then checks if the condition is true. If the condition is not true,
```

pkg/async/cond.go
```diff
@@ -0,0 +1,129 @@
...
+				c.Mu.Unlock()
+			}
+
+			// either way, wait for the next signal
+			waitErr := c.Wait(ctx)
+			if waitErr != nil {
+				return func() {}, waitErr
+			}
+		} else {
+			// condition met, return the unlock function and nil error
```
jkinkead, line 120: Run-on comment:
```suggestion
			// condition met, return the unlock function and nil error.
```

pkg/async/cond.go
```diff
@@ -0,0 +1,129 @@
...
+// unlock is safe to call regardless  of error.
+// Error should only be returned if the context ends before the condition is met.
+//
+// If it returns without error, it also locks the provided locker and the caller must call the returned function
+// to unlock it. Until they call unlock, the state should not be changed.
+//
+// This method encapsulates the instructions in sync.Cond.Wait:
+//
+// """
+// Because c.L is not locked while Wait is waiting, the caller
```
jkinkead, line 74: Update `L` to `Mu` throughout the example?

jkinkead (approved):
Mostly LGTM! One small bug and a couple of comment suggestions.

pkg/async/cond.go
jkinkead: Probably should close `t` in my example before line 35.

pkg/async/cond.go
AndrewWinterman: yah, thanks!

PR #330, opened 2023-08-01
feat: Deprecate async/pool in favor of github.com/sourcegraph/conc/pool

jaredallard:
Deprecates the major methods in `async/pool` in favor of using
`github.com/sourcegraph/conc/pool` instead. The `async/pool` has had
usability issues for a long time now, namely hard to use `Wait()` as
well as almost impossible to get errors out of. The `conc/pool` library
is a much better, more actively maintained libary.

Included is a new `README.md` with instructions on how to migrate all
usecases I could discern from across the `getoutreach` organization.

The main change here is the removal of bufferred pools of work items
that could be dynamically resized as well as rejected if there was no
room. I could only find one repo, `mailroomapi`, that used this
functionality. It is likely that the block until accepted behaviour will
be more acceptable to this service, but also I don't feel that warrants
the additional overhead/code duplication that we own here today.

### DIFF
--- a/pkg/async/pool/pool.go
+++ b/pkg/async/pool/pool.go
@@ -33,7 +33,14 @@ func (of OptionFunc) Apply(opts *Options) {
 // SizeFunc tells the pool whether it should increase or decrease number of workers
 type SizeFunc func() int
 
-// ConstantSize provides
+// ConstantSize provides constant size for the pool.
+//
+// Deprecated: This library is being deprecated in favor of using
+// https://pkg.go.dev/github.com/sourcegraph/conc/pool instead. There is
+// no equivalent because calls to (*Pool).Go() will block if there are
+// no free workers. Control the size of the worker pool by calling
+// (*Pool).WithMaxGoroutines(). For more information, see the README:
+// https://github.com/getoutreach/gobox/tree/main/pkg/async/pool/README.md
 func ConstantSize(size int) OptionFunc {
 	return func(opts *Options) {
 		opts.Size = func() int {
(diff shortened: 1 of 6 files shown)


pkg/async/pool/pool.go
```diff
@@ -276,13 +295,28 @@ type loggingScheduler struct {
...
-// WithLogging creates a scheduler which logs the errors returned from the scheduling as well as executing phase
+// WithLogging creates a scheduler which logs the errors returned from
+// the scheduling as well as executing phase.
+//
+// Deprecated: This library is being deprecated in favor of using
+// https://pkg.go.dev/github.com/sourcegraph/conc/pool instead. There is
+// no replacement for this function. Instead, log on each item when
+// calling (*Pool).Go(). For more information, see the README:
+// https://github.com/getoutreach/gobox/tree/main/pkg/async/pool/README.md
 func WithLogging(name string, s Scheduler) Scheduler {
```
jaredallard, line 320: Nothing at Outreach called this.

pkg/async/pool/timeout.go
```diff
@@ -11,7 +11,14 @@ import (
...
-// WithTimeout creates enqueuer that cancel enqueueing after given timeout
+// WithTimeout creates enqueuer that cancel enqueueing after given
+// timeout
+//
+// Deprecated: This library is being deprecated in favor of using
+// https://pkg.go.dev/github.com/sourcegraph/conc/pool instead. There is
+// no equivalent to this function in the new library. For more
+// information, see the README:
+// https://github.com/getoutreach/gobox/tree/main/pkg/async/pool/README.md
 func WithTimeout(timeout time.Duration, scheduler Scheduler) Scheduler {
```
jaredallard, line 22: Nothing at Outreach called this besides one of our `argocd` microservices. Didn't seem worth keeping, but also doesn't work in the new "always block if not available model".

malept (approved):
Happy to own less code

PR #348, opened 2023-08-17
fix(log): ensure logging calls never cause process exit

AndrewWinterman:

The gobox logger cannot handle Infinite values in logs.

Postgres can hold infinite values.

Smartstore CDC tries to log cdc values at debug level, so they’re available in case of error in cdc.

therefore: smartstore cdc tries to log infinity, hits a log.Fatal in gobox and exits immediately, preventing cdc from advancing.

Slack message from Anton Audzei in #qss-support

Offending log.Fatal:

https://github.com/getoutreach/gobox/blob/51f0f1c06538a0c9baacf0c8d3df52b3bd636e86/pkg/log/log.go#L168

## Jira ID

[DT-3972]

## Notes for your reviewers

[DT-3972]: https://outreach-io.atlassian.net/browse/DT-3972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


malept:
The PR title is funny, but a more descriptive title might be:

> fix(log): avoid exiting early with unmarshal errors

AndrewWinterman:
> The PR title is funny, but a more descriptive title might be:
> 
> > fix(log): avoid exiting early with unmarshal errors

Is this a blocker?

pkg/log/log.go
```diff
@@ -165,7 +165,8 @@ func format(msg, level string, ts time.Time, appInfo Marshaler, mm Many) string
 
 	var b bytes.Buffer
 	if err := json.NewEncoder(&b).Encode(entry); err != nil {
-		log.Fatal(err)
+		log.Printf("could not encode %v; %s", entry, err.Error())
```
jaredallard, line 168: Perhaps, instead of calling the standard logger, it would make more sense to just write to `b` with the raw contents of the entry? 

Also, I know this is nitpicky, but it would still be hard to determine where this is coming from right now with this change. I think we should denote that somewhere in the log message.

pkg/log/log_test.go
```diff
@@ -6,12 +6,24 @@ import (
...
+	"testing"
 
 	"github.com/getoutreach/gobox/pkg/app"
 	"github.com/getoutreach/gobox/pkg/log"
 	"github.com/getoutreach/gobox/pkg/log/logtest"
 )
 
+func TestUnmarshalableValues(t *testing.T) {
+	t.Run("Infinity", func(t *testing.T) {
+		log.Info(context.Background(), "infinity is not fine but not a problem", log.F{"party": math.Inf(1)})
```
jaredallard, line 19: Writing to `b` should allow us to test that the value was written in an expected format as well.

jaredallard:
> > The PR title is funny, but a more descriptive title might be:
> > > fix(log): avoid exiting early with unmarshal errors
> 
> Is this a blocker?

Changing the PR title makes sense, merging this fix won't get it out to consumers without a considerable amount of updating across the org, so we can't exactly rush this specific fix out right now.

pkg/log/log.go
AndrewWinterman: Sounds reasonable. Just so I’m clear, do you mean the results of calling fmt.Sprintf(%v)?

jaredallard: Something like `b.WriteString(fmt.Sprint(entry))`, but with my last comment more like:

```go
b.WriteString(fmt.Sprintf("gobox/log: failed to json encode log entry %v; err=%v", entry, err))
```

jaredallard: It's also _highly_ possible that `%v` will be almost useless here, so we _may_ end up wanting to do something like `spew.Sdump(entry)` instead... but we can figure that out after this goes out IMO.

nissim-natanov-outreach: I would also highly recommend adding the type of the entry to whatever log printed, e.g.
`fmt.Sprintf("gobox/log: failed to json encode log entry %v of type %T; err=%v", entry, entry, err`

gplusplus314: We could do an expensive call for a stack trace and write it into the buffer?

AndrewWinterman: yeah, i think i'll leave out the stack, since that seems to have some implications i dont really want to think about right now, and at some point we're gonna collide with the logging rework effort

pkg/log/log.go
```diff
@@ -165,7 +164,17 @@ func format(msg, level string, ts time.Time, appInfo Marshaler, mm Many) string
 
 	var b bytes.Buffer
 	if err := json.NewEncoder(&b).Encode(entry); err != nil {
-		log.Fatal(err)
+		// at this point we need to report the serialization error.
+		// do it in a JSON object so parsers have a better chance of understanding it
+		err = json.NewEncoder(&b).Encode(map[string]string{
+			"message": fmt.Sprintf("gobox/log: failed to JSON encode log entry %v of type %T; err=%v", entry,
```
jkinkead, line 170: How much do we rely on the log serializer to omit sensitive data? It's pretty worrying to serialize an entire object blindly, given that it could contain sensitive info.

What about just logging the type and the error?

AndrewWinterman: into it

AndrewWinterman: fixed

pkg/log/log_test.go
AndrewWinterman: done.

pkg/log/log.go
AndrewWinterman: done.

jkinkead (approved):

pkg/log/log.go
jaredallard: Great catch.

jaredallard (approved):

PR #349, opened 2023-08-21
feat: add capability to have a status code for client canceled.

ChrisPRobinson:
This happens when a server already cancels a request. This will enable better metrics tracking. There were cases where due to a context being canceled an InternalServerError returned. This would cause alerts to trigger when in fact the only issue was that the context was canceled.

## Jira ID

[STAR-1788]

## Notes for your reviewers
This was an issue that was happening for marathon. By the time the server goes to make the request the context has been canceled. At this point its best to simply return "ClientCanceled"

More discussion came from this slack thread.
https://outreach-hq.slack.com/archives/CN9MU7GLW/p1691689393028399

This PR is simply the first part of what was already discussed.

[STAR-1788]: https://outreach-io.atlassian.net/browse/STAR-1788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/statuscodes/statuscodes.go
+++ b/pkg/statuscodes/statuscodes.go
@@ -47,6 +47,10 @@ const (
 	// are flooding the server.  It is expected that the client will back off for some duration and then try again.
 	// Well-behaved services will even return an expected duration for the client to retry-after.
 	RateLimited StatusCode = 705
+	// ClientCanceled is for when the client has canceled the request, and the server has not yet started processing.
+	// It is expectedc that sometimes the client will simply have a canceled request and no server request will have
+	// been made.
+	ClientCanceled StatusCode = 706
 
 	// The 800-swath is for Server-caused error responses
 	// InternalServerError is for when something otherwise uncategorizable has blown up inside the service logic.
@@ -122,6 +126,8 @@ func FromString(s string) (StatusCode, bool) {
 		return Conflict, true
 	case RateLimited.String():
 		return RateLimited, true
+	case ClientCanceled.String():
+		return ClientCanceled, true
 	case InternalServerError.String():
 		return InternalServerError, true
 	case NotImplemented.String():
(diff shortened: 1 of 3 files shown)


pkg/statuscodes/statuscodes.go
```diff
@@ -47,6 +47,10 @@ const (
 	// are flooding the server.  It is expected that the client will back off for some duration and then try again.
 	// Well-behaved services will even return an expected duration for the client to retry-after.
 	RateLimited StatusCode = 705
+	// ClientCanceled is for when the client has canceled the request, and the server has not yet started processing.
+	// It is expectedc that sometimes the client will simply have a canceled request and no server request will have
+	// been made.
```
jaredallard, line 52:
```suggestion
	// ClientCanceled is for when the client has canceled the request, and the server has not yet started processing.
	// It is expected that sometimes the client will simply have a canceled request and no server request will have
	// been made.
```

Also, "the server has not yet started processing" isn't entirely correct, right? A lot of time the client might close the connection mid-request.

We might also want to document the intended usage of this status code since it will be a little difficult to understand for most people to determine how to use this (since the client has hanged up at this point, so how does one respond? for example.)

ChrisPRobinson: @jaredallard I was cherry picking the case that I ran into which is the context indicates that its done. In that case I ran into, I would expect that I would simply open a new connection and retry. 

In the case where part of the request actually occurred and was cut off, I'm honestly not sure what would need to be done there. I can't think of a general action that could be take, nor how it would be figured out how to recover. WDYT

ChrisPRobinson: Actually the more I think about it, the more I would just surface the error to the client with that grpc status code and they can choose their next step.

jaredallard (approved):

deregtd (approved):
what it says on the tin

PR #356, opened 2023-08-29
feat: ability to override traced call Kind

ChrisPRobinson:
This code is used already in rms. https://github.com/getoutreach/rms/blob/4e0ec20fd5e46d71a9fa861cce71eebf67bc5494/pkg/protoapi/app/dialect/router/hotswap_resolver.go#L74

Need to get this in so that I can update the marathon pkg to the latest to fix a bulk operation change

### DIFF
--- a/pkg/trace/call.go
+++ b/pkg/trace/call.go
@@ -97,8 +97,9 @@ func SetCallError(ctx context.Context, err error) error {
 	return SetCallStatus(ctx, err)
 }
 
-// SetCustomCallKind is meant to alter Kind property of traced call.
-// This can be useful as a service specific dimension to slice e.g. grpc_request_handled metric based on additional context
+// SetCustomCallKind is meant to alter the Kind property of a traced call.
+// This can be useful as a service specific dimension to slice, e.g.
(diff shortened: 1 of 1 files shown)


ChrisPRobinson:
Re-basing the branch

pkg/trace/call.go
```diff
@@ -96,6 +97,15 @@ func SetCallError(ctx context.Context, err error) error {
 	return SetCallStatus(ctx, err)
 }
 
+// SetCustomCallKind is meant to alter Kind property of traced call.
+// This can be useful as a service specific dimension to slice e.g. grpc_request_handled metric based on additional context
```
malept, line 101: Nit: wording
```suggestion
// SetCustomCallKind is meant to alter the Kind property of a traced call.
// This can be useful as a service specific dimension to slice, e.g.
// grpc_request_handled metric based on additional context.
```

malept (approved):
This seems fine modulo the comment wording. I would generally prefer to have tests for this since this is a public library.

zaquestion (approved):

PR #357, opened 2023-08-31
feat(aws): add authorize func that can optionally emit credential provider JSON

malept:

This is basically the business logic needed to create a CLI command to generate AWS credentials, either by writing them to the well-known AWS credentials file, or emitting credential provider compliant JSON.

### DIFF
--- a/pkg/cli/aws/aws.go
+++ b/pkg/cli/aws/aws.go
@@ -63,7 +63,13 @@ const (
 // AuthorizeCredentialsOptions are optional arguments for the
 // AuthorizeCredentials function.
 type AuthorizeCredentialsOptions struct {
-	Force  bool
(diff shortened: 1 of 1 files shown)


kaldorn (approved):
That looks reasonable. How can we test it?

malept:
> That looks reasonable. How can we test it?

I've largely been testing it in dry run mode for okta-aws-cli, and in both modes for saml2aws. Seems to work fine. I'm happy to write some automated tests for at least some failure cases in a follow-up PR. Unfortunately the happy path (which effectively involves trying to log into AWS) doesn't seem like it's very amenable to automated tests, particularly on a public project.

joeysmithoutreach (approved):

PR #364, opened 2023-09-15
fix: DT-4023 Fix tracing.ForceSample + SampleRate

richardlarocque:
[DT-4023](https://outreach-io.atlassian.net/browse/DT-4023)

Updates the implementation of `trace.ForceTracing(ctx)` such that it has
an effect only when its returned `ctx` is used.  The modified `ctx` will
always sample any spans associated with it.  Any requests made with this
`ctx` will send an `X-Force-Trace` header to the receiving service.
However, there will be no impacts outside of this particular `ctx`.

Previously, this method would set an effectively process-global flag and
enable force-trace behaviors process-wide.  This is not what the
function signature suggests it does and probably not what its callers
would expect.

Since we're poking at the `otelForceSampler` code anyway, updates the
decisions returned by `ShouldSample` to include a proper "SampleRate"
attribute.  This will let Honeycomb know what kind of dice roll (if any)
led to this trace being sampled, so it can adjust its estimated counts
accordingly.

[DT-4023]: https://outreach-io.atlassian.net/browse/DT-4023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/trace/force_sampler.go
+++ b/pkg/trace/force_sampler.go
@@ -8,19 +8,32 @@ import (
 	"context"
 
 	"github.com/honeycombio/beeline-go/sample"
+	"go.opentelemetry.io/otel/attribute"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/trace"
 )
+
+var sampleRateAttribute = attribute.Key("SampleRate")
 
 const (
 	fieldForceTrace = "force_trace"
 )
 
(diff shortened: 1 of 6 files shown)


yj-yan (approved):
Looks legit to me. Clear explanation and the code changes does the fix in the description.

jlautman (approved):

PR #365, opened 2023-09-22
feat: DT-4029: Disable call info logging default

richardlarocque:
[DT-4029](https://outreach-io.atlassian.net/browse/DT-4029)

Disables the info logs that would previously occur as part of a
`trace.StartCall`'s succesful completion.  Calls that end in errors have
been and will continue to be logged.

Adds a boolean config option to `trace.yaml` so that services can opt
back in to the old behavior.  We don't want services to fear upgrading
to this version of gobox; this flag will let them upgrade without
experiencing an unpleasant behavior change.

Adds a new call option to explicitly enable info logs for particular
instances of `trace.StartCall`.  We hope that folks that rely on these
logs will rely on this more targeted approach to meet their needs,
rather than using the service-wide `trace.yaml` opt-out.

If you would like to keep your service on the more verbose behavior,
add the following snippet to your service's `.override.jsonnet`:
```
{
  trace_configmap+: {
    data_+:: {
      LogCallsByDefault: true,
    }
  }
}
```

[DT-4029]: https://outreach-io.atlassian.net/browse/DT-4029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/trace/call.go
+++ b/pkg/trace/call.go
@@ -18,6 +18,11 @@ import (
 
 var callTracker = &call.Tracker{}
 
+// logCallsByDefault is a package-global flag that controls whether or not
+// non-error `trace.StartCall` calls get an info log or not.  Its value is set
+// in `setDefaultTracer` and it comes from a value in the config file.
+var logCallsByDefault = false
+
 // StartCall is used to start an internal call. For external calls please
 // use StartExternalCall.
 //
@@ -49,7 +54,17 @@ var callTracker = &call.Tracker{}
 func StartCall(ctx context.Context, cType string, args ...log.Marshaler) context.Context {
 	log.Debug(ctx, fmt.Sprintf("calling: %s", cType), args...)
 
-	ctx = StartSpan(callTracker.StartCall(ctx, cType, args), cType)
+	// Specify the default behavior first in line.  It might be overridden
+	// by later args and that's OK.
+	opts := make([]log.Marshaler, 0, len(args)+1)
+	if logCallsByDefault {
+		opts = append(opts, WithInfoLoggingEnabled())
+	} else {
+		opts = append(opts, WithInfoLoggingDisabled())
+	}
+	opts = append(opts, args...)
+
+	ctx = StartSpan(callTracker.StartCall(ctx, cType, opts), cType)
 	AddInfo(ctx, args...)
 
 	return ctx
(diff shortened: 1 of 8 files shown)


pkg/trace/call.go
```diff
@@ -49,7 +54,17 @@ var callTracker = &call.Tracker{}
 func StartCall(ctx context.Context, cType string, args ...log.Marshaler) context.Context {
 	log.Debug(ctx, fmt.Sprintf("calling: %s", cType), args...)
 
-	ctx = StartSpan(callTracker.StartCall(ctx, cType, args), cType)
+	// Specify the default behavior first in line.  It might be overridden
+	// by later args and that's OK.
+	var opts []log.Marshaler
```
huazhoul-outreach, line 59: suggest change to: `opts := make([]log.Marshaler, 0, len(args)+1)`

huazhoul-outreach (approved):

pkg/trace/call.go
```diff
@@ -18,6 +18,11 @@ import (
 
 var callTracker = &call.Tracker{}
 
+// logCallsByDefault is a package-global flag that controls wheter or not
```
jaredallard, line 21:
```suggestion
// logCallsByDefault is a package-global flag that controls whether or not
```

pkg/trace/call_option.go
```diff
@@ -69,6 +69,17 @@ func WithCallOptions(ctx context.Context, opts CallOptions) {
 //	ctx = trace.StartCall(ctx, "http", trace.WithInfoLoggingDisabled())
 func WithInfoLoggingDisabled() call.Option {
 	return func(c *call.Info) {
-		c.Opts.DisableInfoLogging = true
+		c.Opts.EnableInfoLogging = false
+	}
+}
+
+// WithInfoLoggingEnabled enables info logging on the current call
```
jaredallard, line 76: Maybe note here that based on the global value, this may have no effect?

pkg/trace/tracetest/tracetest.go
```diff
@@ -41,10 +43,11 @@ func NewSpanRecorderWithOptions(options Options) *SpanRecorder {
 	fakeConfig := map[string]interface{}{
 		"OpenTelemetry": map[string]interface{}{
 			"SamplePercent": options.SamplePercent,
-			"Endpoint":      "localhost",
+			"Endpoint":      "blackhole.test",
```
jaredallard, line 46: Smart idea :D

jaredallard (approved):

pkg/trace/tracetest/tracetest.go
richardlarocque: Thank you, ChatGPT.

>Ah, I see. In the context of unit tests where you want to simulate a non-responsive or "black hole" hostname, you have a few options:
> 
> 3. Use the Reserved Domains: The .test, .example, .invalid, and .localhost TLDs are reserved and should not be used on the public internet. For unit tests, you can use something like blackhole.test.

pkg/trace/call_option.go
richardlarocque: Good point, will document.

benmishkanian:
Thank you for this, I used it to reduce dsngerrorarchiver logging by 17 million messages per day 😃

PR #370, opened 2023-10-03
fix: DT-3870 use a new timeout context to shut down tracer

yj-yan:
This PR tries to fix the issue when service is shut down, otel needs the context but the context has been cancelled. Here it creates a new context with 3 seconds timeout for shutting down the trace provider.

## Jira ID

[DT-3870]

### DIFF
--- a/pkg/trace/otel.go
+++ b/pkg/trace/otel.go
@@ -180,9 +180,12 @@ func (t *otelTracer) closeTracer(ctx context.Context) {
 
 	t.tracerProvider.ForceFlush(ctx)
 
-	err := t.tracerProvider.Shutdown(ctx)
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), time.Second*3)
+	defer cancel()
+
+	err := t.tracerProvider.Shutdown(ctxTimeout)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


jaredallard (approved):

pkg/trace/otel.go
```diff
@@ -180,9 +180,12 @@ func (t *otelTracer) closeTracer(ctx context.Context) {
...
 	t.tracerProvider.ForceFlush(ctx)
 
-	err := t.tracerProvider.Shutdown(ctx)
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), time.Second*3)
+	defer cancel()
+
+	err := t.tracerProvider.Shutdown(ctxTimeout)
 	if err != nil {
-		log.Error(ctx, "Unable to stop otel tracer", events.NewErrorInfo(err))
+		log.Error(ctx, "Unable to stop otel tracer within the context timeout", events.NewErrorInfo(err))
```
jkinkead, line 188: Probably should be a warning, if we have those ...

yj-yan: Make sense to me. I changed it to warning.

jkinkead (approved):

PR #382, opened 2023-10-18
fix: make updater.Disabled usable

PeSme:

Updater package defines `Disabled bool` that should be usable as a global kill switch of updater services. Unfortunately it is useless:
- the suggested usage via ldflags does not work. It fails `github.com/getoutreach/gobox/pkg/cli/updater.Disabled: cannot set with -X: not a var of type string (type:bool)` and [documentation](https://pkg.go.dev/cmd/link) is clear - it must be string if we want to use it this way.
- alternatively if you try `updater.Disabled = true` it does not really disable the initialization which likely fails and logs a warning.

So this PR fixes the behavior. Turns it into string and stops `defaultOptions()` successfully.

### DIFF
--- a/pkg/cli/updater/updater.go
+++ b/pkg/cli/updater/updater.go
@@ -41,10 +41,10 @@ import (
 //
 // or you can do it in main() before UseUpdater is called:
 //
-//	updater.Disabled="true"
+//	updater.Disabled = "true"
 //
 // Any value other than "true" is considered false.
-// The type must be string to allow -ldflags "-X ..."
+// The type is string to allow for invoking via `go run -ldflags "-X ..."`.
 var Disabled = "false"
 
 // UseUpdater creates an automatic updater.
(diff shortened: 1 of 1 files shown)


pkg/cli/updater/updater.go
```diff
@@ -33,10 +33,19 @@ import (
...
+// This should usually be done with an ldflag:
+//
+//	go run -ldflags "-X github.com/getoutreach/gobox/pkg/cli/updater.Disabled=true" ...
+//
+// or you can do it in main() before UseUpdater is called:
+//
+//	updater.Disabled="true"
+//
+// Any value other than "true" is considered false.
+// The type must be string to allow -ldflags "-X ..."
```
malept, line 47: Nit: code style, grammar
```suggestion
//	updater.Disabled = "true"
//
// Any value other than "true" is considered false.
// The type is string to allow for invoking via `go run -ldflags "-X ..."`.
```

malept (commented):
This seems fine, but I had some suggestions in the new docs.

malept (approved):

jaredallard (approved):
Thank you for catching this!!

PR #391, opened 2023-10-30
fix(aws): add support for upstream okta-aws-cli v2

malept:

Fixes a regression in #384 and adds support for v2 of okta-aws-cli by checking the output of `okta-aws-cli --version`.

Also updates the call to `okta-aws-cli` to directly set the appropriate `OKTA_*` environment variables from the box config, instead of relying on the parent shell environment. This caused problems when the external dependency that pulled the values from the box config into the shell environment broke on macOS.

## Jira ID

[DT-4091]

[DT-4091]: https://outreach-io.atlassian.net/browse/DT-4091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yj-yan (approved):
LGTM. Hopefully okta-aws-cli future major versions will have the compatibility.

pkg/cli/aws/aws.go
```diff
@@ -208,12 +239,88 @@ func refreshCredsViaOktaAWSCLI(ctx context.Context, copts *CredentialOptions, ac
...
+	}
+
+	output, err := io.ReadAll(stdout)
+	if err != nil {
+		return false, errors.Wrap(err, "could not read version output")
+	}
+
+	if err := cmd.Wait(); err != nil {
+		return false, errors.Wrap(err, "problem completing execution of okta-aws-cli --version")
+	}
```
jaredallard, line 295: Why not use [`cmd.Output`](https://pkg.go.dev/os/exec#Cmd.Output)?

pkg/cli/aws/aws.go
```diff
@@ -208,12 +239,88 @@ func refreshCredsViaOktaAWSCLI(ctx context.Context, copts *CredentialOptions, ac
 	return nil
 }
 
-// runCmd is a wrapper for running a command via exec.CommandContext
-// and passing through stdin/stdout/stderr.
-func runCmd(ctx context.Context, name string, args ...string) error {
-	cmd := exec.CommandContext(ctx, name, args...)
+// oktaAwsCLICmd is a convenience function that creates an exec.Cmd
+// instance for okta-aws-cli.
```
jaredallard, line 243:
```suggestion
// instance for the okta-aws-cli.
```

pkg/box/box.go
```diff
@@ -42,6 +42,19 @@ type AWSConfig struct {
 	// Known values:
 	// * okta-aws-cli (default)
 	RefreshMethod string `yaml:"refreshMethod"`
+
+	Okta OktaConfig `yaml:"okta"`
```
jaredallard, line 46:
```suggestion
  // Okta contains configuration for using Okta authenication
  // to authenticate with AWS.
	Okta OktaConfig `yaml:"okta"`
```

pkg/cli/aws/aws_test.go
```diff
@@ -161,17 +167,54 @@ func Test_refreshCredsViaOktaAWSCLI(t *testing.T) {
...
 		msg := hook.LastEntry().Message
 		assert.Assert(t, strings.HasPrefix(msg, "Dry Run: okta-aws-cli"))
 		assert.Assert(t, !strings.Contains(msg, "--write-aws-credentials"))
-		assert.Assert(t, strings.Contains(msg, "--format credential-provider"))
+		assert.Assert(t, cmp.Contains(msg, "--format process-credentials"))
 	})
 }
+
+func Test_oktaAwsCliVersionOutputMatchesV1(t *testing.T) {
+	t.Run("version 1 matches", func(t *testing.T) {
```
jaredallard, line 193: nit: If you're going to use sub-tests, it would be preferable to use [table tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests)

pkg/cli/aws/aws.go
malept: I disagree with this suggestion, it sounds awkward putting a definite article in front of a command name.

pkg/cli/aws/aws_test.go
malept: Replaced with non-sub-tests

pkg/cli/aws/aws.go
malept: Because I didn't know it existed :sweat_smile:

pkg/cli/aws/aws_test.go
```diff
@@ -190,28 +190,26 @@ func Test_refreshCredsViaOktaAWSCLI(t *testing.T) {
 }
 
 func Test_oktaAwsCliVersionOutputMatchesV1(t *testing.T) {
```
jaredallard, line 192: I would've preferred a table test, to be honest. But, that's something we can always change later :D

jaredallard (approved):

PR #393, opened 2023-11-01
feat: olog package

jaredallard:
Example implementation of a wrapper around the `slog` stdlib
package added in `go 1.21`. This wrapper focuses on keeping
the `slog.Logger` struct as the core logger, but adding the
ability to control log levels of loggers globally. This is done
by adding a dynamic `slog.Leveler` that is aware of where
a logger was created (package and module). There is a
global registry that can change the log level based on those
same dimensions.

Also included is an example of a typed logger that is built
on top of the `slog.Logger` type for things that should
follow a specific format. In this case, audit logs.

It's important to note this is a reference implementation, not
an actual one. Loggers should be created per-package and
stored as a global variable, or passed around within that
package, based on the consumers preferences. Note that
loggers will also be attributed to the package that created
them and that is an intentional decision.


pkg/olog/audit.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2023 Outreach Corporation. All Rights Reserved.
+
+// Description: Contains a strongly-typed logger for audit logs using
+// slog.
+//
+// Note: This is an EXAMPLE of an extension to the slog.Logger, it is
+// not meant to actually be used.
```
jaredallard, line 7: Important: This shows what an extension to the logger would look like if someone wanted to support logging structured logs for a specific thing (e.g., `audit` logs)

pkg/olog/olog.go
```diff
@@ -0,0 +1,36 @@
...
+// New creates a new slog instance that can be used for logging. The
+// provided logger use the global handler provided by this package. See
+// the documentation on the 'handler' global for more information.
+//
+// The logger will be automatically associated with the module and
+// package that it was instantiated in. This is done by looking at the
+// call stack.
+func New() *slog.Logger {
+	// Look up the call stack to find the module and package.
+	// Create a new handler and store it in the map.
```
jaredallard, line 24: This is a TODO, just showing that those values would be changed.

pkg/olog/default_handler.go
```diff
@@ -0,0 +1,85 @@
...
+const (
+	JSONHandler DefaultHandlerType = iota
+	TextHandler
+)
+
+// determineDefaultHandler sets the default handler based on the current
+// environment. If the `os.Stderr` is a TTY, then the default handler
+// is a `slog.TextHandler`. Otherwise, the default handler is the
+// `slog.JSONHandler`.
+func determineDefaultHandler() {
```
jaredallard, line 42: This will allow us to automatically pretty-print logs when running in development.

pkg/olog/log_level_registry.go
jaredallard, line 1: A global registry of handlers is required to enable controlling log-levels of handlers dynamically.

pkg/olog/log_level.go
```diff
@@ -0,0 +1,67 @@
...
+func newLeveler(lr *levelRegistry, addrs []string) slog.Leveler {
+	return &leveler{lr, addrs}
+}
+
+// SetGlobalLevel sets the global logging level used by all loggers by
+// default that do not have a level set in the level registry. This
+// impacts loggers that have previously been created as well as loggers
+// that will be created in the future.
+func SetGlobalLevel(l slog.Level) {
+	level.Store(int64(l))
```
zaquestion, line 66: Seems like something we could wrap in a once? Seemingly an application should only ever call this once and a subsequent or dynamic usage of this would be inappropriate?

PR #411, opened 2023-11-15
feat: Add Pointer Package

romain-tadiello:

This PR aims to add a pointer package to be used across our services and librairies. 
I had to do pointer conversion in API-Proxy [here](https://github.com/getoutreach/apiproxy/blob/f3707830d6d25f94057182027037c709d7a14c4d/internal/apiv2/openapi.go#L33) and saw `ptr.SomeType()` in several places (not using generic). 
As we are using gql federation in our services we will surely need more pointer conversion and I would like to have the possibility to use generics.
- Added tests with different types including some that are not supported by [github.com/aws/smithy-go/blob/main/ptr](https://github.com/aws/smithy-go/blob/main/ptr)

## Jira ID

No Jira, just an idea of package that would be nice to have

### DIFF
--- a/pkg/pointer/pointer.go
+++ b/pkg/pointer/pointer.go
@@ -11,10 +11,12 @@
 // with map / slice
 package pointer
 
+// ToPtr returns a pointer to {object} of type *T
 func ToPtr[T any](object T) *T {
 	return &object
 }
 
+// ToSlicePtr returns a slice of pointers of type *T pointing to each element of {objects}
 func ToSlicePtr[T any](objects []T) []*T {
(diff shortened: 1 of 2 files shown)


pkg/pointer/pointer.go
jaredallard, line 1: All exported functions should have comments on them, please!

romain-tadiello: Done 👍

jaredallard (approved):

antonaudzei-outreach (approved):

PR #417, opened 2023-12-06
fix(trace): Support parent-based trace sampling (DT-4122)

richardlarocque:
[DT-4122](https://outreach-io.atlassian.net/browse/DT-4122)

Updates the gobox trace library to support OpenTelemetry parent-based
sampling.  Now, when a Go service receives a traceparent header that
includes the "sampled" bit, indicating that the parent sampled this
trace, then the Go service will also sample this trace.

This contrasts with the previous way of doing things, inherited from
Honeycomb beeline libraries, where each service independently computed
a deterministic hash of the trace ID and used that to make independent
but convergent sampling decisions.

The motivation for this change is the fact that Gazelle/Giraffe, and
later the flagship, have hopped on board this more standard parent-based
sampling convention.  We've been seeing broken traces as the Go services
lag behind and make different sampling decisions based on different
algorithms.

This change is complicated by the fact that we want to retain support
for Honeycomb's "sample rate" attribute, which auto-adjusts the counts
and sums in their UI to account for sampling.  There is no finalized
OpenTelemetry standard for this.  We implement our own solution, using
the (cross-service spanning) "tracestate" attribute to carry information
about the sample rate that was used to make the sampling decision.  We
don't expect cross-service propagation of this attribute to work *yet*,
but the pieces are there that we could attempt this in the future.

Breaks the existing sampler into its component parts, defining new and
separate samplers for the "honeycomb deterministic sampling" and
"x-force-trace" and "hack to attach sample rate attributes".  These
samplers are more unit-testable than the all-in-one samplers that
preceded them.  We also provide one `defaultSampler` that ties together
all these pieces, and we make it the one and only option for gobox OTel
tracing for now.

[DT-4122]: https://outreach-io.atlassian.net/browse/DT-4122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


richardlarocque:
Note that there are two medium-sized commits in this PR.

The first one has the sampler-related changes described in the PR description.

The second removes support for Honeycomb-style trace propagation headers.  I'll have more to say about that in some self-comments.

pkg/trace/roundtripper_test.go
```diff
@@ -141,124 +141,6 @@ func TestRoundtripper(t *testing.T) {
...
-						"spanContext.traceID":     traceID,
-						"spanKind":                "internal",
-						"startTime":               differs.AnyString(),
-						"SampleRate":              int64(1),
-					},
-				}
-			},
-		},
-		{
-			name:                  "OtelServerHoneycombClient",
```
richardlarocque, line 212: I'd started this PR with just the first commit.  It was almost ready to go, but then this test failed.

I could have just deleted this test.  It's not like we care about clients using Honeycomb-style headers these days, right?  But when I dug a bit more I decided it might be best to rip the band-aid off and drop support for Honecyomb-style header propagation entirely.

pkg/trace/roundtripper_test.go
```diff
@@ -320,22 +203,6 @@ func otelCallRoundTripper(t *testing.T, state *roundtripperState) []map[string]i
 	return state.Ended()
 }
 
-func honeycombInitRoundTripperState(t *testing.T) *roundtripperState {
-	recorder := tracetest.NewSpanRecorder()
-	server := httptest.NewServer(
-		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			prop, err := propagation.UnmarshalHoneycombTraceContext(r.Header.Get(propagation.TracePropagationHTTPHeader))
```
richardlarocque, line 327: I'm not sure why anyone would write an HTTP handler that digs directly into trace headers like this.  That's the framework's job, right?

After I ripped out the support for Honeycomb headers this started to fail pretty badly, and for good reason.  The `"X-Honeycomb-Trace"` header had disappeared.

I'm going to assume this is a test-only issue.  It should be only framework-y things that look at these headers.

pkg/trace/propagation.go
```diff
@@ -138,16 +91,6 @@ func (t *otelTracer) contextFromHeaders(ctx context.Context, hdrs map[string][]s
...
 	}
 
-	if header.Get(OtelPropagationHeader) == "" {
-		prop, err := propagation.UnmarshalHoneycombTraceContext(header.Get(propagation.TracePropagationHTTPHeader))
-		if err == nil {
-			_, headers := propagation.MarshalW3CTraceContext(ctx, prop)
-			for k, v := range headers {
-				header.Set(k, v)
-			}
-		}
```
richardlarocque, line 148: This is the problem.

This block of code notices that there is no OpenTelemetry/W3C trace headers, then goes looking for Honeycomb headers.  If it finds them, it "converts" whatever it finds to W3C headers then hacks those re-exported values onto the request context so the true OpenTelemetry header parsing code further up the stack can use it.

The problem is that the honeycomb headers don't have a "sampled" flag like OpenTelemetry headers do.  They don't support parent-based sampling, or at least not in this way.  The resulting OpenTelemetry headers have the sampled flag hard-coded to "false", which means we never sample any of the traces based on requests that take this branch.

pkg/trace/propagation.go
```diff
@@ -111,20 +78,6 @@ func (t *otelTracer) toHeaders(ctx context.Context) map[string][]string {
...
-
-	_, prop, err := propagation.UnmarshalW3CTraceContext(ctx, headers)
-	if err != nil {
-		return result
-	}
-
-	result.Set(
-		propagation.TracePropagationHTTPHeader,
-		propagation.MarshalHoneycombTraceContext(prop))
-
```
richardlarocque, line 127: As far as I can tell, this block of code:
- Serializes the current trace context to the W3C header format.
- Uses honeycomb beeline code to parse those headers into structures the beeline libs understand.
- Re-serializes this beeline-formatted data into Honeycomb-style headers.
- Sets those Honeycomb headers in the return value.

In other words, it's a hack to ensure we forward HC-style as well as OpenTelemetry-style headers.  This seems unnecessary if we're dropping support for receiving those headers, and we think that pretty much everyone has been ignoring the HC-style headers for quite some time now.  Hence this deletion.

richardlarocque:
Some devenv test links:
- Can confirm [sample rate adjustment](https://ui.honeycomb.io/outreach-a0/datasets/dev/result/HvLBUrF7pkj) in RMS service.  This was a case of "assume the default because remote parent was sampled".
-

richardlarocque:
I hooked this up to `whoami` and ran some tests, "SampleRate" was hard-coded to 25%:
```
# Trace starts in `whoami`, sampled at 1/4, tagged with appropriate "Sample Rate".
# https://ui.honeycomb.io/outreach-a0/datasets/dev/result/BbhotBpztXT
grpcurl -plaintext localhost:5000 api.Whoami/ForwardThenWhoami

# X-Force-Trace ensures the request is always sampled, effective "Sample Rate" is 1, though HC does not show this explicitly.
# https://ui.honeycomb.io/outreach-a0/datasets/dev/result/hykLSkSTUc7
grpcurl -plaintext -H "X-Force-Trace: true" localhost:5000  api.Whoami/ForwardThenWhoami

# Pretend this is a non-sampled remote trace.  It is never sampled.
grpcurl -plaintext -H 'traceparent: 00-80e1afed08e019fc1110464cfa66635c-7a085853722dc6d2-00'

# Pretend this is a sampled remote trace.  It assumes remote's sample rate is the same as its own, and tags accordingly.
# https://ui.honeycomb.io/outreach-a0/datasets/dev/result/zay6ZieJiCb/trace/pXhj3AeKmyt?fields[]=s_name&fields[]=s_serviceName&span=c8e67b3992c3bf85
# (this trace is confused because of remnants from the previous test,
# where mint did sometimes sample,
# because I haven't hacked my local mint into supporting parent-based sampling yet.)
grpcurl -plaintext -H 'traceparent: 00-80e1afed08e019fc1110464cfa66635c-7a085853722dc6d2-01'
```

george-e-shaw-iv (approved):
This seems good and from my perspective doesn't modify the public interface to this package or break any existing workflows past assumptions already called out within the PR.

pkg/trace/sampling.go
```diff
@@ -0,0 +1,285 @@
...
+// guess as to the upstream sample rate when we continue from a remote parent.
+//
+// The provided sample rate is the number of requests seen per requset sampled.
+func defaultSampler(sampleRate uint) sdktrace.Sampler {
+	parentSampler := sdktrace.ParentBased(
+		// new, non-remote trace: use HC deterministic sampler.
+		NewHoneycombDeterministicSampler(sampleRate),
+
+		// Assume our parent uses the same sample rate that we do.  It's
+		// an odd assumption, but it's the better options aren't fully
```
jaredallard, line 62: This reads a little oddly, I feel like it's missing something.

pkg/trace/sampling.go
```diff
@@ -0,0 +1,285 @@
...
+//
+// [1] https://www.w3.org/TR/trace-context/
+const sampleRateKey = "sample_rate"
+
+// SetTraceStateSampleRate sets the sample rate in the tracestate.
+func SetTraceStateSampleRate(ts trace.TraceState, sampleRate uint) trace.TraceState {
+	updated, err := ts.Insert(sampleRateKey, fmt.Sprintf("%d", sampleRate))
+	if err != nil {
+		// This really shouldn't happen.
+		log.Warn(context.TODO(), "failed to insert sample rate into tracestate", events.NewErrorInfo(err))
```
jaredallard, line 268: nit: This is a huge nit, but I reserve `context.TODO()` for cases where we plan on updating the signature to include a context. This doesn't seem like something we would go add it to (if we did, we would do it now, right?). So, I would say we should use `context.Background` instead.

pkg/trace/sampling.go
```diff
@@ -132,7 +133,7 @@ func (s *forceTraceHeaderSampler) ShouldSample(p sdktrace.SamplingParameters) sd
 func NewHoneycombDeterministicSampler(sampleRate uint) sdktrace.Sampler {
 	sampler, err := sample.NewDeterministicSampler(sampleRate)
 	if err != nil {
-		panic(err)
+		panic(fmt.Errorf("failed to create deterministic sampler: %w", err))
```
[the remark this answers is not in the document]
richardlarocque: The linter thought that it was wrong to pass in an error as `%v` and that's why I changed it.

If it leads to better panic messages, that's great.  If not, well, it should still kill the program, right?

jaredallard:
@richardlarocque Do you plan on merging this soon? I'd like to avoid incorporating the last commit into my own branches to help avoid a merge conflict (but it's not worries if you're unable to get this in soon!)

richardlarocque:
> @richardlarocque Do you plan on merging this soon? I'd like to avoid incorporating the last commit into my own branches to help avoid a merge conflict (but it's not worries if you're unable to get this in soon!)

Hi @jaredallard !

Yep, almost there.  For some reason the changes I've made have affected whitespace in the stringification of histogram metrics, in a way that broke unit tests.  It's behaving kinda weird.  I think it has something to do with protobuf relfection magic deep inside the prometheus packages.

I don't feel like spending a lot of time on this at right this moment.  As far as I can tell, all the metric _values_ are as they should be, so I don't think this is a real-world correctness issue.  My formatting change in the latest commit seems to have fixed it.

I'm going to push up one more change with a comment to explain what I've done and why, and then I'm good to merge.

richardlarocque:
I rebased and now the spacing behavior has shifted back again.  :(

I'll try again to fix.

richardlarocque:
Almost there...

I tried to sweep this under the rug but I kept getting burned by it.  I adjusted the single-spacing and then Circle still flaked; apparently now it's gone back to wanting to see double-spaced outputs.

I'm going to try to put an end to this madness.  I removed the stringification and string compare logic and replace it with something that does fancier, well-typed `cmp.Equal` checks on the protobufs.  It's uglier, but it should be more stable.  At the very least I expect future failures to be more meaningful.

Will merge once I get this new thing to pass.

PR #423, opened 2023-12-11
feat(ometrics): adds the ometrics package

jaredallard:
This adds a `ometrics` package (as a sub-module, more on that later)
aimed at configuring the global otel.MeterProvider. This allows
consumers to instrument metrics through the standard otel metrics
package. The ideal usage is below:

```go
// main.go
ometrics.InitializeMeterProvider(ctx, ExporterTypePrometheus)

// every single package that exports metrics.
var meter otel.GetMeterProvider().Meter("packageName")

func doAThing() {
  counter, err := meter.Float64Counter("my_counter", api.WithDescription("counts a thing"))
  if err != nil {}
  counter.Add(ctx, 5)
}
```

**Note**: Yes, there is not a notion of package global metrics like
Prometheus had. Metrics have to be instrumented in a function and then
passed along to whatever consumes them. That was a design decision by
otel maintainers.

There's some open questions in comments in the code that I was unable to
get to as well as future direction. For example, to fully support the
collector exporter (push) there needs to be configuration instrumented.
It's up to the future implementor to decide how that should be done.
Though, I would suggest not utilizing environment variables for it :)

This is added a sub-module because the otel dependencies have been
_very_ painful in the past, so I'd like to avoid proliferating them to
every project that consumes `gobox`. Otherwise, just sourcing `pkg/app`
can break an application today.


pkg/ometrics/ometrics.go
```diff
@@ -0,0 +1,103 @@
...
+
+	// ExporterTypeCollector exports metrics to the otel collector. This
+	// is akin to the "push" model of metrics, for those familiar with
+	// the prometheus model.
+	ExporterTypeCollector
+)
+
+// InitializeMeterProvider initializes the global meter provider to be
+// backed by the provided exporter.
+func InitializeMeterProvider(ctx context.Context, t ExporterType, opts ...Option) error {
```
jaredallard, line 45: I would note that there technically could be an import requirement for this. Examples show the meter provider _just working_ when used as a package global, but I'm dubious and haven't looked at the actual `GetMeterProvider().Meter` code to determine if that's the case or not.

amhester: My initial thought looking at their code was that meters created before the global provider was set would use whatever provider they were created with. However, after looking through the code more for setting the global provider, it seems they may have some means of updating the provider on pre-existing meters through some delegate concept. Would need to look even more deep into it, but it at least seems like order SHOULDN'T matter here. I will likely need to update my docs as well to account for that.

amhester:
One quick comment on this PR after looking at this section of the [design brief](https://outreach-io.atlassian.net/wiki/spaces/DT/pages/2853765388/Go+Observability+Tooling+Design#Migration-Strategy.2), is that I'm not sure there's currently a great way to delineate between the metrics provided by this new package and the metrics provided by existing prometheus instrumentation (assuming we continue using prometheus). We currently seem to expose metrics via http using the httpx package which is included in all stencil-golang projects by default. The httpx package uses promhttp to setup the metrics handler on the `/metrics` endpoint. The promhttp package I believe will automatically collect metrics written by the otel prometheus exporter as well. Not sure if this is really particularly important, but figured I'd call it out.

Also, I could be missing something and this may not be an issue.

pkg/ometrics/ometrics.go
```diff
@@ -0,0 +1,103 @@
...
+	case ExporterTypePrometheus:
+		reader, err = prometheus.New()
+	case ExporterTypeCollector:
+		// TODO(jaredallard): We'll want to plumb in gRPC configuration as
+		// well as potentially the periodic read config here.
+		var exporter *otlpmetricgrpc.Exporter
+		exporter, err = otlpmetricgrpc.New(ctx)
+		if err == nil {
+			reader = sdkmetric.NewPeriodicReader(exporter, sdkmetric.WithInterval(c.Collector.Interval))
+		}
```
amhester, line 82: From what I could tell, and I am very new to this lib, but I believe it's worth discussing how we want to expose the configuration and creation of the gRPC client required for this collector type. They have this example [here](https://github.com/open-telemetry/opentelemetry-go/blob/main/example/otel-collector/main.go) for the otlp trace collector, which should be conceptually the same as for the otlp metric collector, and it seems the gRPC client is required. However, tbf, I haven't looked too deeply into this.

amhester (approved):

pbr0ck3r (approved):

pkg/ometrics/ometrics.go
yj-yan: Good call out, seemingly we could pass in the gRPC Options as an argument to the function.

yj-yan (approved):

pkg/ometrics/ometrics.go
```diff
@@ -0,0 +1,103 @@
...
+	default:
+		return fmt.Errorf("exporter type provided is unknown")
+	}
+	if err != nil {
+		return fmt.Errorf("failed to create exporter: %w", err)
```
malept, line 87: suggestion: unclear why `errors.Wrap` wasn't used here.
```suggestion
		return errors.Wrap(err, "failed to create exporter")
```

malept (commented):
I largely have docs suggestions.

pkg/ometrics/ometrics.go
jaredallard: I think I introduced this, I've largely moved away from `errors.Wrap` in favor of `%w` because it's the idiomatic way to wrap errors now.

PR #472, opened 2024-03-06
ci: restencil for CI fix

malept:

Fixes the CI failure due to the kubectl apt change.


yj-yan (approved):

malept:
E2E failure is an existing failure, should be fixed in a separate PR.

PR #476, opened 2024-03-11
fix: stop dual initiation of trace clients

kaldorn:

If the otel collector endpoint is specified we should _not_ initiation a new client for honeycomb.

## Jira ID

[DT-4286]

## Notes for your reviewers

[DT-4286]: https://outreach-io.atlassian.net/browse/DT-4286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/trace/otel.go
+++ b/pkg/trace/otel.go
@@ -92,42 +92,19 @@ func (t *otelTracer) initTracer(ctx context.Context, serviceName string) error {
 	mp := noop.NewMeterProvider()
 	otel.SetMeterProvider(mp)
 
-	var exp *otlptrace.Exporter
+	var client otlptrace.Client
 
-	if t.Otel.CollectorEndpoint == "" {
-		key, err := t.Otel.APIKey.Data(ctx)
-		if err != nil {
-			log.Error(ctx, "Unable to fetch otel API key", events.NewErrorInfo(err))
-		}
-
-		headers := map[string]string{
-			"x-honeycomb-team":    strings.TrimSpace(string(key)),
-			"x-honeycomb-dataset": t.Otel.Dataset,
-		}
-
-		client := otlptracegrpc.NewClient(
-			otlptracegrpc.WithEndpoint(t.Otel.Endpoint+":443"),
-			otlptracegrpc.WithHeaders(headers),
-		)
-
-		exp, err = otlptrace.New(ctx, client)
-		if err != nil {
-			log.Error(ctx, "Unable to start trace exporter", events.NewErrorInfo(err))
-		}
-
+	// We want to default to initializating and sending traces through the opentelemetry collectors.
+	// But the fallthrough is to send to Honeycomb directly.
+	if t.Otel.CollectorEndpoint != "" {
+		client = t.newOpentelemetryClient(ctx)
 	} else {
-		var err error
-		log.Info(ctx, fmt.Sprintf("Using OTEL Collector Endpoint: %s", t.Otel.CollectorEndpoint))
-		client := otlptracegrpc.NewClient(
-			otlptracegrpc.WithEndpoint(t.Otel.CollectorEndpoint),
-			// There is no need for TLS because we're sending traffic to a kubernetes service
-			otlptracegrpc.WithInsecure(),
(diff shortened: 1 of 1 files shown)


pkg/trace/otel.go
```diff
@@ -92,24 +92,42 @@ func (t *otelTracer) initTracer(ctx context.Context, serviceName string) error {
...
+		)
+
+		exp, err = otlptrace.New(ctx, client)
+		if err != nil {
+			log.Error(ctx, "Unable to start trace exporter", events.NewErrorInfo(err))
+		}
+
+	} else {
+		var err error
+		log.Info(ctx, fmt.Sprintf("Using OTEL Collector Endpoint: %s", t.Otel.CollectorEndpoint))
```
richardlarocque, line 120: Do we need this log?  Maybe we can just omit it?

yj-yan (approved):
Changes look good to me, stop sending trace to honeycomb twice.

pkg/trace/otel.go
```diff
@@ -92,24 +92,42 @@ func (t *otelTracer) initTracer(ctx context.Context, serviceName string) error {
...
-	)
+		headers := map[string]string{
+			"x-honeycomb-team":    strings.TrimSpace(string(key)),
+			"x-honeycomb-dataset": t.Otel.Dataset,
+		}
 
-	exp, err := otlptrace.New(ctx, client)
-	if err != nil {
-		log.Error(ctx, "Unable to start trace exporter", events.NewErrorInfo(err))
+		client := otlptracegrpc.NewClient(
```
richardlarocque, line 108: I wonder if it might be cleaner to take the client construction out of here?

We could have something separate that builds the `otlptracegrpc.NewClient` and/or its options.  We could even have two, one for `Endpoint` vs. `CollectorEndpoint`, if that works out better.

Very optional, though.

pkg/trace/otel.go
kaldorn: If I added that, it'd be a debug log for sure. I'll remove it. I'm 100% in favor of removing logs after things work.

pkg/trace/otel.go
kaldorn: I think that makes sense. I'm not tied to the if/else here because I think it's cluttered.

pkg/trace/otel.go
```diff
@@ -92,24 +92,42 @@ func (t *otelTracer) initTracer(ctx context.Context, serviceName string) error {
...
+			otlptracegrpc.WithHeaders(headers),
+		)
+
+		exp, err = otlptrace.New(ctx, client)
+		if err != nil {
+			log.Error(ctx, "Unable to start trace exporter", events.NewErrorInfo(err))
+		}
+
+	} else {
+		var err error
```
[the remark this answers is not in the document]
kaldorn: Ah, good suggestion.

pkg/trace/otel.go
kaldorn: I've attempted to clean this up.

pkg/trace/otel.go
```diff
@@ -92,21 +92,16 @@ func (t *otelTracer) initTracer(ctx context.Context, serviceName string) error {
...
-	key, err := t.Otel.APIKey.Data(ctx)
-	if err != nil {
-		log.Error(ctx, "Unable to fetch otel API key", events.NewErrorInfo(err))
-	}
+	var client otlptrace.Client
 
-	headers := map[string]string{
-		"x-honeycomb-team":    strings.TrimSpace(string(key)),
-		"x-honeycomb-dataset": t.Otel.Dataset,
+	// We want to default to initializating and sending traces through the opentelemetry collectors.
```
malept, line 97: Nit: wording
```suggestion
	// We want to default to initialize and send traces through the OpenTelemetry collectors.
```

malept (approved):
Looks legit, just need to fix a comment.

richardlarocque (approved):
LGTM.

PR #522, opened 2024-05-23
fix(cli): always allow --skip-update even when updater is disabled

PeSme:

When updater is disabled (via -ldflags -X, for local builds, or because initialization failed) it would be nice to still allow `--skip-updates` parameter which is hardcoded on many places calling various commands normally relying on updater.

Of course, `--skip-updates` has no effect in that case, but the command will not fail complaining about
```
Incorrect Usage: flag provided but not defined: -skip-update
```


pkg/cli/updater/updater.go
```diff
@@ -153,6 +154,7 @@ func (u *updater) defaultOptions() error {
 		var err error
 		u.executablePath, err = exec.ResolveExecuable(os.Args[0])
 		if err != nil {
+			u.hookSkipUpdateIntoCLI()
```
yj-yan, line 157: Question: is there a use case to pass `--skip-updates` flag when returning error?

PeSme: I'm not sure about this one on the line 157. But for sure there it makes sense for the one on line 167. And I did not want to make a difference, the skip update parameter simply should be always accepted. The error is logged but does not prevent execution anyway.

yj-yan (approved):
The change looks legit to me, passing the skip-updates flag to avoid missing flag error.

PR #525, opened 2024-05-28
feat(olog): add charm-style formatter for logrus

asms:
Add a charm-style formatter for logrus to have consistent logging formats.

### Background

`log/slog` has been widely accepted by the community as the new standard for logging in Go. This olog package is our implementation of it which uses the charm logger. However, we are currently using logrus in many places for CLI applications. In order to have consistency in our log formats, I decided to create a custom logrus formatter which matches the text format of the charm logger.

### Notes

Most of this is copied from the [charm text formatter](https://github.com/charmbracelet/log/blob/82b5630d2e68c2cf4c972a926be90149fe0c60b9/text.go), but it's been slightly tweaked. The `Format` method is what fulfills the logrus interface. It creates the right request shape and byte buffer to call into `textFormatter`. The bytes are then extracted and returned to logrus.

While this mimics the formatting of olog, it still can't handle switching to JSON when stdout is not attached to a TTY. I don't intend to nor do I think we should handle that.

### Example

image

### DIFF
--- a/pkg/olog/logrus.go
+++ b/pkg/olog/logrus.go
@@ -8,8 +8,8 @@
 //
 //              Example usage:
 //              ```l := logrus.New()
-//				l.SetFormatter(NewCharmTextFormatter())
-//				l.SetReportCaller(true)```
+//              l.SetFormatter(NewCharmTextFormatter())
+//              l.SetReportCaller(true)```
 //
 //              Example output:
 //              `14:03:38 INFO <qss/qss.go:77> message key=value`
--- a/pkg/olog/logrus.go
+++ b/pkg/olog/logrus.go
@@ -57,10 +57,9 @@ func NewCharmTextFormatter() logrus.Formatter {
 	}
 }
 
-// Format implements logrus.Formatter using a charm-style text format
-func (l *logrusCharmTextFormat) Format(entry *logrus.Entry) ([]byte, error) {
+func logrusToCharmLevel(logrusLevel logrus.Level) charm.Level {
 	var level charm.Level
-	switch entry.Level {
+	switch logrusLevel {
 	case logrus.FatalLevel, logrus.PanicLevel:
 		level = charm.FatalLevel
 	case logrus.ErrorLevel:
@@ -72,6 +71,12 @@ func (l *logrusCharmTextFormat) Format(entry *logrus.Entry) ([]byte, error) {
 	case logrus.DebugLevel, logrus.TraceLevel:
 		level = charm.DebugLevel
 	}
+	return level
+}
+
+// Format implements logrus.Formatter using a charm-style text format
+func (l *logrusCharmTextFormat) Format(entry *logrus.Entry) ([]byte, error) {
+	level := logrusToCharmLevel(entry.Level)
 	entries := []interface{}{
 		charm.TimestampKey, entry.Time,
 		charm.LevelKey, level,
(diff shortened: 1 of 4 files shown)


pkg/olog/logrus.go
```diff
@@ -0,0 +1,359 @@
...
+//              Intended to be used by applications which currently use logrus and want consistently formatted
+//              while using a mix of loggers (logrus and olog). This code was mostly copied from [here](1).
+//
+//              [1]: https://github.com/charmbracelet/log/blob/82b5630d2e68c2cf4c972a926be90149fe0c60b9/text.go "Charm Text Format"
+//
+//              Example usage:
+//              ```l := logrus.New()
+//				l.SetFormatter(NewCharmTextFormatter())
+//				l.SetReportCaller(true)```
+//
```
yj-yan, line 13: nit: extra spaces
```suggestion
//              ```l := logrus.New()
//		   l.SetFormatter(NewCharmTextFormatter())
//		   l.SetReportCaller(true)```
//
```

pkg/olog/logrus_test.go
```diff
@@ -0,0 +1,141 @@
...
+		{
+			name:     "simple message",
+			expected: "INFO info\n",
+			msg:      "info",
+			fields:   nil,
+			level:    logrus.InfoLevel,
+		},
+		{
+			name:     "ignored message",
+			expected: "",
```
yj-yan, line 35: Question: Why this debug log is expected to be ignored?

yj-yan (commented):
Thanks for making the contribution and the comprehensive test cases. 

I take it as this formatter is intended make the current CLI which uses `logrus` to be compatible with the `olog` `TextHandler`. Since DT is in progress of migrating our CLI to olog, I am not sure if there will be caveats of the logging format. Overall I think the implementation looks good for now. 

I am okay with the change, but I am holding to also let other DT member review it.

pkg/olog/logrus_test.go
asms: This was copied over from the charm logger tests. I think it's because the default log level is Info, so debug logs are suppressed. This led me to realize that logrus puts the responsibility of suppressing logs on the formatter. I've added that, so it will now suppress logs below the logger's log level.

pkg/olog/logrus.go
asms: I think some tabs ended up in there creating inconsistent spacing because it displayed fine in vscode.

pkg/olog/logrus.go
```diff
@@ -0,0 +1,369 @@
...
+//              [1]: https://github.com/charmbracelet/log/blob/82b5630d2e68c2cf4c972a926be90149fe0c60b9/text.go "Charm Text Format"
+//
+//              Example usage:
+//              ```l := logrus.New()
+//              l.SetFormatter(NewCharmTextFormatter())
+//              l.SetReportCaller(true)```
+//
+//              Example output:
+//              `14:03:38 INFO <qss/qss.go:77> message key=value`
```
malept, line 15: suggestion: this makes more sense to have this documentation associated with `NewCharmTextFormatter` since it's unlikely for someone to read the file doc comments, and more likely for someone to read the function's documentation.

malept (approved):
This seems fine to merge after all of the documentation related comments are addressed.

yj-yan (approved):

PR #548, opened 2024-07-15
fix(olog): add lock to SetOutput

yj-yan:

Make the SetOutput function thread safe.

## Jira ID

[XX-XX]

### DIFF
--- a/pkg/olog/olog.go
+++ b/pkg/olog/olog.go
@@ -16,8 +16,13 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"sync"
 
 	"github.com/getoutreach/gobox/pkg/callerinfo"
+)
+
+var (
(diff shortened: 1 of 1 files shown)


pkg/olog/olog.go
```diff
@@ -118,6 +123,8 @@ func NewWithHandler(h slog.Handler) *slog.Logger {
 
 // SetOutput sets the global logger output to desired writer.
```
malept, line 124: suggestion: the docs should be updated to mention that there's a mutex on the global logger output.

malept (changes requested):

pkg/olog/olog.go
yj-yan: Thanks. Updated the comment.

malept (approved):

PR #553, opened 2024-07-24
feat(services): run function

AndrewWinterman:

This PR extracts most of the `main` function provided by stencil-golang into a function clients can call instead.

QSS runs a number of peripheral services-- CDC, Ingest, Polaroid, SmartWorker, etc., each of which need to set up their own main function. To do this, they basically copy whats in stencil-golang. Wouldn't it be better if we could use a library function instead? I think it would be, since it eases the burden when it comes time to update any service-wide behavior.

Of course, we'll still want stencil to provide the kubernetes objects, e.g. trace.yaml for example, and my hope is we'll stencil in depending on this method instead of just doing what it does currently inline.

## Jira ID

[XX-XX]


evadnoob:
I'm curious why this change is in gobox, is that the correct location, or should it be in in `services`?

pkg/run/run.go
```diff
@@ -0,0 +1,157 @@
...
+// optionFunc is a function that implements serviceOption
+type optionFunc func(*runOpts) error
+
+// apply implements serviceOption.
+func (s optionFunc) apply(o *runOpts) error {
+	return s(o)
+}
+
+// OptLogger sets the logger. Otherwise it defaults to olog.New
+func OptLogger(l *slog.Logger) Option {
```
george-e-shaw-iv, line 53: nit: the idiom I've see across the company and in the ecosystem for functions returning functional options are to prefix them with `With`. So `WithLogger`, `WithHTTPAppHandler`, etc

AndrewWinterman: I think that is fine too, I have a mild preference for Opt, although really I should make the interface an Opt too.

AndrewWinterman:
> I'm curious why this change is in gobox, is that the correct location, or should it be in in `services`?
> 
> 

I think it should either be in services or stencil golang. I put it here at first because I was worried about import circles, but as I worked it out that turned out to be unfounded

pkg/run/run.go
```diff
@@ -0,0 +1,157 @@
...
+	options ...Option,
+) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// deafult options
+	opts := &runOpts{
+		httpAppHandler: http.NotFoundHandler(),
+		log:            olog.New(),
+		httpAddr:       "127.0.0.1:5000",
```
george-e-shaw-iv, line 113: this should either be one or two exported constants probably

pkg/run/run.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	for _, o := range options {
+		err := o.apply(opts)
+		if err != nil {
```
george-e-shaw-iv, line 118:
```suggestion
		if err := o.apply(opts); err != nil {
```

pkg/run/run.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/getoutreach/stencil-golang/pkg/serviceactivities/automemlimit"
+	"github.com/getoutreach/stencil-golang/pkg/serviceactivities/gomaxprocs"
+	"github.com/getoutreach/stencil-golang/pkg/serviceactivities/shutdown"
+
+	"github.com/getoutreach/httpx/pkg/handlers"
+)
+
+// runOpts are the options to pass to Run to configure your service
+type runOpts struct {
+	httpAppHandler http.Handler
```
george-e-shaw-iv, line 33: I think I know why you added this as a static member of the runners, but I'm still not sure it's the right move in the spirit of abstracting our service activity running framework into a library. I think I'd rather just have people to inject this as a generic runner opposed to force this as a default activity.

AndrewWinterman: I did it because of metrics, liveness check, readiness check, etc. to a large degree, you don't have a first-class service without this thing

george-e-shaw-iv: Right and that is what I figured, but even as you have it right now it doesn't even provide that. You still have to inject the right one in. That is why I think we should just leverage the generic `runners` field for this too, but I'm ok either way

AndrewWinterman: > You still have to inject the right one in.

From what I can tell, it should. My test hits the liveness endpoint... or do you mean something else?

AndrewWinterman: Otoh, this prevents adding custom logic to the liveness/healthcheck

george-e-shaw-iv: Talked on slack - this was a misunderstanding on my part

pkg/run/run.go
george-e-shaw-iv: Haha I don't really mind either way. Here are some hand wave-y metrics on usage here between the two if they make you think any differently - https://github.com/search?q=org%3Agetoutreach%20path%3A*.go%20%22func%20With%22&type=code
https://github.com/search?q=org%3Agetoutreach+path%3A*.go+%22func+Opt%22&type=code

george-e-shaw-iv (approved):

AndrewWinterman:
porting this to services.

PR #556, opened 2024-07-31
feat:  add logger interface to be compatible with logrus and slog

yj-yan:

This is an implementation of compatible layer of log and olog package.

## Jira ID

[XX-XX]


pkg/log/logger.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+// Logger is the compatible log interface
+type Logger struct {
+	logrusLogger *logrus.Logger
+	ologLogger   *slog.Logger
+	useLogrus    bool
+}
+
+func New(useLogrus bool) *Logger {
```
malept, line 23: issue: this should probably be named something else, otherwise it gets confusing as to why you need to use this versus all of the other module functions.

pkg/log/logger.go
```diff
@@ -0,0 +1,82 @@
...
+		// add attr field to log message
+		log.ologLogger.DebugContext(ctx, message, "attr", m)
+	}
+}
+
+func (log *Logger) Info(ctx context.Context, message string, m Marshaler) {
+	if log.useLogrus {
+		Info(ctx, message, m)
+	} else {
+		log.ologLogger.InfoContext(ctx, message, "attr", m)
```
malept, line 45: question: how do marshaled attributes look in Datadog when using `pkg/log` vs `pkg/olog` when using this struct?

pkg/log/logger.go
```diff
@@ -0,0 +1,82 @@
...
+func New(useLogrus bool) *Logger {
+	if useLogrus {
+		l := logrus.New()
+		return &Logger{logrusLogger: l, useLogrus: true}
+	}
+	l := olog.New()
+	return &Logger{ologLogger: l, useLogrus: false}
+}
+
+func (log *Logger) Debug(ctx context.Context, message string, m Marshaler) {
```
malept, line 32: question: The current use of `pkg/log`'s logging methods allow multiple `m Marshaler`, and it's used a lot that way in services, e.g.

```go
log.Error(ctx, "Something bad happened", err, log.F{"user.email", "hello@example.com"})
```

How will this need to be migrated over?

malept (changes requested):
Can you add docs to all of the new functions/methods?

Additionally, the struct will need migration docs and possibly examples on how to migrate.

malept:
We're not doing this. We might as well just have folks migrate directly to `olog`, rather than migrate to this interface.

PR #603, opened 2024-11-14
feat(aws): add debug flags

malept:

Adding support to set `--debug` and `--debug-api-calls` may be helpful when debugging strange issues with `okta-aws-cli`.


ekalinichev (approved):
❤️

PR #606, opened 2025-01-08
fix: Upgrade go-git to fix CVE

lelandbatey:

This change updates the go-git library from v5.12.0 to v5.13.1 in order to stop the wiz_cli_image_scan from complaining about CVE-2025-21613:

https://github.com/advisories/GHSA-v725-9546-7q7m

Note that the go-git library is only being used for its INI parsing code and thus this CVE isn't even relevant, but we're updating anyway to stop the scanner from complaining.

## Jira ID

No jira, I had this happen to me and hold up my PR, so I'm fixing it.


ChrisPRobinson (approved):
LGTM

PR #632, opened 2025-03-14
feat: numeric conversions package

KaliCZ:

A package for numeric conversions in case you cannot change the types to match. The package gives an underflow/overflow error and it is type-safe since it creates an interface for numeric types only.

## Jira ID

[XX-XX]

## Notes for your reviewers
There's a lot of room for improvements for the tests. But I don't feel that it would be so detrimental that it should be blocking.


HyeonJi-Kang (approved):
LGTM

ekalinichev (changes requested):
Discussed, gobox is not a good place for it

PR #634, opened 2025-03-18
feat(olog): finish registry

AndrewWinterman:
## What this PR does

Completes the `olog` log-level registry so log levels can actually be *set*, not just respected:

1. `SetLevel(level, address...)` - programmatic way to set the level for a module/package address (or the global default).
2. `ConfigureFromFile(path)`  - loads a YAML level config from disk once 
3. `PollConfigurationFile(ctx, path, interval, errFunc)` - watch said yaml file for changes, which means with appropriate kubernetes mounting changes to log level can be applied without restart (see https://oneuptime.com/blog/post/2026-02-02-kubernetes-configmap-hot-reload/view for someone else doing this)
5. `SetLevelResolver(r LevelResolver)` - optional, context-aware hook that lets client code override the level per log emission. 

## Why

Jira: [DT-4401] — *olog: missing configuration for log level*. `olog` honored registry levels internally, but there was no supported way for a service to change them: no exported setter, no on-disk configuration, and no way to plug in dynamic sources (e.g. feature flags) without forking gobox.

Concretely this unblocks:
- Turning on debug logging for one noisy package in production without a redeploy (edit the config file / ConfigMap, poller picks it up).
- Turning off noisy logs for a third party that also uses this package.
- Per-request or per-org log level overrides driven by feature flags, with gobox staying dependency-free.

## How it works

- **Registry setter** (`pkg/olog/log_level.go`): `SetLevel` writes into the existing address-keyed level registry, so handlers already holding a `slog.Leveler` observe the change immediately — no handler re-creation.
- **File config** (`pkg/olog/config.go`): a small `Config` struct maps address → level string via `stringLevel`. Unknown level strings are skipped rather than silently applied as the zero value (`INFO`). `PollConfigurationFile` polls on an interval, tracks `lastModTime` so it only reloads when the file actually changes, and routes errors to a caller-supplied `errFunc` that decides whether to keep polling. Cancellation is honored through a ctx-aware `sleep` (deliberately not `pkg/async` — that would create an `olog → async → log → olog` import cycle).
- **Level resolver** (`pkg/olog/level_resolver.go`): `resolverHandler` wraps the inner handler and consults a global `atomic.Pointer[LevelResolver]` in `Enabled`/`Handle`, passing the handler's addresses and the registry-derived level. If no resolver is registered (the default), the cost is a single nil check; if one returns `(level, false)` the registry level wins. This keeps feature-flag integration entirely in client code.
- Fixtures + tests cover file loading, unknown levels, mtime-gated polling, and resolver overrides. `pkg/olog/README.md` documents all three entry points.

## Notes for reviewers

- `LevelResolver` runs on the logging hot path. Adapters should be cheap/cached — don't call a remote flag service inside it.

[DT-4401]: https://outreach-io.atlassian.net/browse/DT-4401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Impact on `pkg/log`

`pkg/log` is unchanged by this PR (no files touched, no API added or removed), but it is `olog`'s biggest consumer, so the new registry surface reaches it indirectly:

**Only the slog path is affected.** `pkg/log` builds its logger via `olog.New()` in `setupSlog()`/`SetShouldUseSlog()`, which is only used when `GOBOX_AS_SLOG_FACADE` is set (or `SetShouldUseSlog(true)`/`SetHandler` is called). The vintage formatter path (`format()` + `Output()`) does not go through `olog` handlers at all, so `SetLevel`, `ConfigureFromFile`, and `SetLevelResolver` have **no effect** on services still on the legacy writer. `log.Debug`'s existing buffer-until-error behavior is likewise untouched.

**Address gotcha: fixed — `log.*` now resolves against the real call site.** `olog.New()` decides a logger's registry identity by walking the stack *at construction time* (`runtime.Callers` → `runtime.CallersFrames` → package/module strings). `pkg/log` constructs its logger once, in `setupSlog()`, from inside `pkg/log` — so every `log.Info/Warn/Error` in every service inherited `pkg/log`'s identity:

- package address `github.com/getoutreach/gobox/pkg/log`
- module address `github.com/getoutreach/gobox`

A config entry for a service's own package could not change the level of its own `log.Debug` calls. Construction site ≠ call site, and the registry was keyed on the former.

The lever already existed: `slogIt` was already calling `runtime.Callers(3, pcs[:])` to populate `source` on the record, then discarding the PC. The fix reuses it as the registry key:

1. `callerinfo.GetCallerInfoFromPC(pc uintptr)` — `GetCallerInfo` already cached PC → `CallerInfo` (`moduleLookupByPC`), with the stack walk only producing the key. Split at that seam; the frame-resolution half is now a shared helper. `GetCallerInfo`'s signature and behavior are unchanged.
2. `olog.NewForPC(pc uintptr)` — same body as `New()`, metadata derived from a PC instead of a stack walk, routed through the unchanged `createHandler` so leveler, `resolverHandler`, and `module`/`modulever` attrs behave identically. It does not panic on a bad PC (a runtime condition, not a programming error); it degrades to the immediate caller, then the main module.
3. `pkg/log` caches one logger per resolved package path in a `*sync.Map`. Amortized cost per emission is one map load. Pointer, not value, because the reset paths (`setupSlog`, `SetShouldUseSlog`, `SetOutput`) reassign it and copying a lock-bearing struct trips `copylocks`.

This could not be done inside a single handler: `slog.Handler` splits the decision between `Enabled(ctx, level)` — the gate, which receives no `Record` and therefore no PC — and `Handle(ctx, r)`, by which point filtering has already happened. Identity must be bound to the handler instance, so it is one handler per caller package.

Test: `pkg/log/internal/callerpkg` exists only to call `log.Debug` from its own package path, and `TestLevelRoutesToCallerPackage` asserts both directions (level on `pkg/log` must *not* enable it; level on `callerpkg` must). Verified non-vacuous — with `pkg/log/log.go` reverted the test fails with `pkg/log address enabled DEBUG for caller package`.

**`SetHandler` opts out entirely.** `log.SetHandler(h)` routes through `olog.NewWithHandler`, which skips `createHandler` — no leveler, no `resolverHandler`. Services installing an OTel log bridge via `SetHandler` get none of the new registry/file/resolver behavior and must do their own level gating.

**Hot-path cost on the facade.** `createHandler` now wraps its handler in `resolverHandler`, so each `log.*` emission on the slog path adds one `atomic.Pointer` load in `Enabled` before delegating. With no resolver installed (the default) that is the entire added cost; with one installed, the resolver runs on every emission from `pkg/log` for the whole process.

**Output format is stable.** `default_handler.go` changes are confined to keeping `charmLevelHandler` in sync with a dynamic leveler; JSON output (what log pipelines actually ingest) is byte-identical, and TTY output remains charm-rendered as on `main`.

**No import-cycle risk.** `olog` still does not import `pkg/log`; the config poller deliberately uses a local ctx-aware `sleep` instead of `pkg/async` to keep it that way.


pkg/olog/README.md
```diff
@@ -358,6 +358,48 @@ func SetGlobalLevel(l slog.Level)
 
 SetGlobalLevel sets the global logging level used by all loggers by default that do not have a level set in the level registry. This impacts loggers that have previously been created as well as loggers that will be created in the future.
 
+## [func SetLevel(l slog.Level, address ...string)](https://github.com/getoutreach/gobox/blob/main/pkg/olog/log_level.go#71)
+
+```go
+func SetLevel(l slog.Level, address ...string)
+```
+
+`SetLevel` SetLevel sets the log-level for the provided addresses, which are modules or file paths
```
malept: Nit: punctuation
```suggestion
`SetLevel` SetLevel sets the log level for the provided addresses, which are modules or file paths.
```

pkg/olog/log_level.go
```diff
@@ -65,3 +67,8 @@ func newLeveler(lr *levelRegistry, addrs []string) slog.Leveler {
 func SetGlobalLevel(l slog.Level) {
 	level.Store(int64(l))
 }
+
+// SetLevel sets the log-level for the provided addresses, which are modules or file paths
```
malept: Nit: see above
```suggestion
// SetLevel sets the log level for the provided addresses, which are modules or file paths.
```

pkg/olog/README.md
```diff
@@ -358,6 +358,48 @@ func SetGlobalLevel(l slog.Level)
...
+`ConfigureFromFile` loads the level configuration from the provided path.
+
+Configuration looks like:
+
+```yaml
+log:
+  - level: WARN
+    address: github.com/getoutreach/gobox/pkg/olog
+  - level: ERROR
+    address: github.com/getoutreach/smartstore
```
malept: Nit: use a different (public) repo

```suggestion
    address: github.com/getoutreach/goql
```

pkg/olog/log_level.go
```diff
@@ -40,10 +40,12 @@ type leveler struct {
...
 // determined using the following logic:
 //
-//   - If the global level is set, then that level is returned.
-//   - Otherwise, the level is retrieved from configuration using the
+//   - The level is retrieved from configuration using the
 //     config package. The configuration is automatically reloaded when
 //     changes are detected.
+//   - If no address level can be retreived, and the global level is set, then
+//     that level is returned.
+//   - otheriwse the level is Info.
```
malept: Nit: typo
```suggestion
//   - Otherwise, the level is `Info`.
```

pkg/olog/log_level.go
```diff
@@ -40,10 +40,12 @@ type leveler struct {
 // handler to determine which logs should be printed. The level is
 // determined using the following logic:
 //
-//   - If the global level is set, then that level is returned.
-//   - Otherwise, the level is retrieved from configuration using the
+//   - The level is retrieved from configuration using the
 //     config package. The configuration is automatically reloaded when
 //     changes are detected.
+//   - If no address level can be retreived, and the global level is set, then
```
malept: Nit: typo
```suggestion
//   - If no address level can be retrieved and the global level is set, then
```

kaldorn (approved):

pkg/olog/config.go
```diff
@@ -0,0 +1,109 @@
...
+	for ctx.Err() == nil {
+		stat, err := os.Stat(logCfgFilePath)
+
+		if err != nil {
+			ok := fn(fmt.Errorf("failed to stat %s: %w", logCfgFilePath, err))
+			if !ok {
+				return
+			}
+			continue
+		}
```
malept, line 102: issue (blocking): Stat-error path skips the poll interval, causing a busy loop

When `os.Stat` fails (e.g. a mounted ConfigMap hasn't appeared yet at pod startup), this branch calls `fn` and then `continue`s, which skips the `sleep(ctx, pollInterval)` call at the end of the loop. The poller then spins calling `os.Stat` with no delay until the file appears or `fn` returns false.

```suggestion
		if err != nil {
			ok := fn(fmt.Errorf("failed to stat %s: %w", logCfgFilePath, err))
			if !ok {
				return
			}
			sleep(ctx, pollInterval)
			continue
		}
```

---
_Generated by [Claude Code](https://claude.ai/code)_

pkg/log/log.go
```diff
@@ -212,11 +223,30 @@ func slogIt(ctx context.Context, lvl slog.Level, message string, m []Marshaler)
...
+			pkgKey = "unknown"
+		}
+		if cached, ok := callerLoggers.Load(pkgKey); ok {
+			handler = cached.(*slog.Logger).Handler()
+		} else {
+			l := olog.NewForPC(pcs[0])
+			actual, _ := callerLoggers.LoadOrStore(pkgKey, l)
+			handler = actual.(*slog.Logger).Handler()
+		}
+	}
```
malept: issue (blocking): `callerLoggers` is read outside `slogLock` while it's reassigned under `slogLock` elsewhere

`slogIt` reads the package-level `callerLoggers` variable at lines 241 and 245 without holding `slogLock`, but `setupSlog` (line 89) and `SetShouldUseSlog` (line 107) reassign it to a new `*sync.Map` while holding `slogLock`. That's an unsynchronized read/write on the same variable, which `go test -race` would flag with a concurrent `SetOutput`/`SetShouldUseSlog` call racing against logging. It is not just a race-detector nit: a goroutine that misses the reassignment keeps using a stale cached `*slog.Logger` bound to the old output writer, so `SetOutput` can silently fail to redirect some in-flight loggers.

Consider reading `callerLoggers` inside the same `slogLock` critical section as `isCustom`/`defaultLog` above it, or replacing the plain package variable with an `atomic.Pointer[sync.Map]`.

---
_Generated by [Claude Code](https://claude.ai/code)_

pkg/olog/level_resolver.go
```diff
@@ -0,0 +1,88 @@
...
+			// charmlog.Logger filters again in Handle using its own
+			// level, so keep it in sync with the resolved level.
+			if c, isCharm := h.inner.(*charmLevelHandler); isCharm {
+				c.inner.SetLevel(charmlog.Level(resolved))
+			}
+			return level >= resolved
+		}
+	}
+	return h.inner.Enabled(ctx, level)
+}
```
malept, line 73: issue (blocking): Resolver-driven level override can race across concurrent goroutines sharing a cached logger

When a `LevelResolver` is installed and the wrapped handler is the charm/Text handler, `Enabled` mutates the shared `charmLevelHandler.inner`'s level (line 65) as a side effect so the later `Handle` call honors the resolved level. `pkg/log` caches one `*slog.Logger` per caller package and shares it across every goroutine logging from that package, so two concurrent calls with different context-resolved levels (the resolver's own use case — e.g. per-request feature flags) can interleave: goroutine A's `Enabled` sets the shared level for A's record, goroutine B's `Enabled` overwrites it before A's `Handle` runs, and A's record is filtered/emitted using B's level instead of its own.

This only affects the charm/Text handler path — `slog.JSONHandler` (used whenever output isn't a TTY) doesn't consult this mutable state, so most production traffic is unaffected. Still worth addressing given the resolver is explicitly documented as safe to call on every emission.

---
_Generated by [Claude Code](https://claude.ai/code)_

pkg/olog/config.go
```diff
@@ -0,0 +1,109 @@
...
+var stringLevel = map[string]slog.Level{
+	slog.LevelDebug.String(): slog.LevelDebug,
+	slog.LevelInfo.String():  slog.LevelInfo,
+	slog.LevelWarn.String():  slog.LevelWarn,
+	slog.LevelError.String(): slog.LevelError,
+	"OFF":                    slog.Level(100),
+}
+
+// Config is level configuration for olog.
+// The address is either a moddule or a package, and the level is one of
```
malept: nitpick: Typo: "moddule" → "module"

```suggestion
// The address is either a module or a package, and the level is one of
```

---
_Generated by [Claude Code](https://claude.ai/code)_

pkg/olog/config.go
```diff
@@ -0,0 +1,109 @@
...
+// DEBUG, INFO, WARN, ERROR, or OFF
+type Config struct {
+	Levels []struct {
+		// module or package path
+		Address string `yaml:"address"`
+		Level   string `yaml:"level"`
+	} `yaml:"olog"`
+}
```
malept, line 42: suggestion (non-blocking): Consider a named type instead of an anonymous struct for `Levels`

`Levels []struct { Address string; Level string }` uses an anonymous struct. `config_test.go:19-22` has to redeclare that same anonymous struct verbatim to build a test literal, which is usually a sign a named type (e.g. `type LevelConfig struct { Address, Level string }`, then `Levels []LevelConfig`) would be easier to reuse and reference.

---
_Generated by [Claude Code](https://claude.ai/code)_

pkg/olog/config.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	for _, configuredLevel := range c.Levels {
+		l, ok := stringLevel[strings.ToUpper(configuredLevel.Level)]
+		if !ok {
+			New().Error("unknown level", "level", configuredLevel.Level, "address", configuredLevel.Address)
+			continue
+		}
+		globalLevelRegistry.Set(l, configuredLevel.Address)
+	}
```
malept, line 75: question (non-blocking): Should unknown levels be surfaced through the returned error, not just logged?

`ConfigureFromFile` logs unknown levels via `New().Error(...)` and otherwise returns `nil`. A caller driving this through `PollConfigurationFile`'s `fn(err)` callback never learns that a specific entry was skipped — only a full YAML-parse failure surfaces as an error. Is that intentional (apply valid entries, rely on log output alone for bad ones), or should unknown-level entries also be aggregated into the returned error so callers can alert on them?

---
_Generated by [Claude Code](https://claude.ai/code)_

go.mod
```diff
@@ -47,6 +47,7 @@ require (
 	golang.org/x/sync v0.22.0
 	golang.org/x/term v0.45.0
 	golang.org/x/tools v0.49.0
+	gopkg.in/yaml.v3 v3.0.1
```
malept: suggestion: this library has been archived, use the maintained one instead (which is already in `go.mod`), `go.yaml.in/yaml/v3`:
```suggestion
```

pkg/olog/config.go
```diff
@@ -0,0 +1,109 @@
...
+// whichever comes first. Inlined (instead of pkg/async.Sleep) to avoid
+// an import cycle: async -> log -> olog.
+func sleep(ctx context.Context, d time.Duration) {
+	t := time.NewTimer(d)
+	defer t.Stop()
+	select {
+	case <-ctx.Done():
+	case <-t.C:
+	}
+}
```
malept, line 127: thought(non-blocking): this is unfortunate because in the near future we will be adding retry functions to `gobox/pkg/async` and it seems like this could simplify this greatly.

malept (changes requested):

pkg/log/log.go
```diff
@@ -70,6 +70,12 @@ var (
 	// log is a structured logger instance.
 	log *slog.Logger
 
+	// customHandlerInstalled indicates whether a custom handler was installed via SetHandler.
+	customHandlerInstalled bool
+
+	// callerLoggers caches *slog.Logger instances by caller package path.
+	callerLoggers = &sync.Map{}
```
AndrewWinterman: rename this to callerLoggersMap for clarity

pkg/olog/config.go
AndrewWinterman: probably, and then the Poller can discover it too

pkg/olog/config.go
AndrewWinterman: can always pull that in when it lands.

malept (approved):

PR #696, opened 2025-07-03
fix(cli): regression related to cleanup/exit functions

reseul:
We detected the issue since a consuming tool (`devenv`) stopped returning `1` on error.
`defer` does nothing with the returned `func()`, dev has to call that further.

## Jira ID

[XX-XX]


malept (approved):
Looks legit, sorry about that

huazhoul-outreach (approved):

PR #740, opened 2025-10-19
feat: gobox as slog facade

AndrewWinterman:

## What
This PR adds an option to use slog for log output for all gobox logs.

Among other things, slog offers a logfmt format which is more readable for local development, as well as more conventional leveling infrastructure.

I did not preserve the debug buffer behavior, but I think that would actually be relatively straightforward at this point, so we can add it back in.

## Why
QSS has a few CLIs that sometimes run in production and sometimes run locally (migrations, topic creator, probably more)

It would be great to have logfmt locally, since it offers signficant readability improvements for local dev.

My hope is that this would help cut down on #dev-tooling-support requests that basically have a resolution in finding a needle of useful error output in a haystack of useless logs.

## Jira ID

[XX-XX]

## Notes for your reviewers

Startup of a service running with `devenv apps run -t .`

```
09:30:22 INFO <tombstone/config.go:104> Configuration data of the application:
 module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc HTTPPort=8000 ListenHost="" grpc.ConnectionTimeout=10s grpc.DefaultUnaryRequestTimeout=10s grpc.GracefulStopTimeout=15s grpc.KeepAliveMinTime=30s grpc.KeepAliveTime=5s grpc.KeepAliveTimeout=30s grpc.MaxConnectionAge=5m0s grpc.MaxConnectionAgeGrace=10s grpc.MaxConnectionIdle=2m0s grpc.MaxRecvMsgSize=10000 grpc.MaxSendMsgSize=10000 grpc.Port=5005 redis="{redis redis.tombstone--bento1a.svc.cluster.local:26379 redacted   0 leader 0 0 0 0 false}" smartstore="{2 1}" tombstone.v1="&{persit {168h0m0s}}"
09:30:22 INFO <tombstone/main.go:80> starting module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc HTTPPort=8000 ListenHost="" app.name=tombstone app.pid=689 app.version=v1.6.0-18-ga548e8c deployment.namespace=tombstone--bento1a grpc.ConnectionTimeout=10s grpc.DefaultUnaryRequestTimeout=10s grpc.GracefulStopTimeout=15s grpc.KeepAliveMinTime=30s grpc.KeepAliveTime=5s grpc.KeepAliveTimeout=30s grpc.MaxConnectionAge=5m0s grpc.MaxConnectionAgeGrace=10s grpc.MaxConnectionIdle=2m0s grpc.MaxRecvMsgSize=10000 grpc.MaxSendMsgSize=10000 grpc.Port=5005 redis="{redis redis.tombstone--bento1a.svc.cluster.local:26379 redacted   0 leader 0 0 0 0 false}" service_name=tombstone smartstore="{2 1}" tombstone.v1="&{persit {168h0m0s}}"
09:30:22 INFO <settings/schema_settings.go:92> loading smartstore schema settings module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc smartstore.resource_name=tombstone smartstore.resource_type=PostgreSQL
09:30:22 INFO <settings/schema_settings.go:101> bytes from schema settings module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc
  bytes=
  │ [
  │  {
  │   "dirName": "",
  │   "name": "tombstone",
  │   "target": "Org"
  │  }
  │ ]
09:30:22 INFO <settings/schema_settings.go:113> loaded schema settings module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc filename=pg-tombstone-schemas.yaml smartstore.resource_name=tombstone smartstore.resource_type=PostgreSQL smartstore.schemasettings.schemas_0="{tombstone tombstone Org  /run/config/outreach.io/migrations}"
09:30:22 INFO <settings/settings.go:59> loading smartstore settings module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc smartstore.resource_name=tombstone smartstore.resource_type=PostgreSQL
09:30:22 INFO <settings/settings.go:66> loaded smartstore settings module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc filename=pg-tombstone-config.yaml shards=map[default:0x40005e33e0] smartstore.database_name=tombstone smartstore.resource_name=tombstone smartstore.resource_type=PostgreSQL smartstore.users="map[infra:{pg-tombstone-users {} {}} readonly:{pg-tombstone-users {} {}} readwrite:{pg-tombstone-users {} {}}]"
09:30:22 INFO <settings/settings.go:59> loading smartstore settings module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc smartstore.resource_name=tombstone smartstore.resource_type=PostgreSQL
09:30:22 INFO <settings/settings.go:66> loaded smartstore settings module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc filename=pg-tombstone-config.yaml shards=map[default:0x40005e3760] smartstore.database_name=tombstone smartstore.resource_name=tombstone smartstore.resource_type=PostgreSQL smartstore.users="map[infra:{pg-tombstone-users {} {}} readonly:{pg-tombstone-users {} {}} readwrite:{pg-tombstone-users {} {}}]"
09:30:22 INFO <log/with.go:31> smartstore.sharding: initialized org shard cache module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc smartstore.org_shard_schema_cache.size=2 smartstore.resource_name=tombstone smartstore.schema_name=tombstone timing.service_time=0.001181042
09:30:22 INFO <log/with.go:31> smartstore.sharding: initialized org shard cache module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc smartstore.org_shard_schema_cache.size=2 smartstore.resource_name=tombstone smartstore.schema_name=tombstone timing.service_time=0.000283959
09:30:22 INFO <smartstore/conn_pool_router.go:322> smartstore: creating rw connection pool for shard module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc smartstore.shard_key=default
09:30:22 INFO <smartstore/conn_pool_router.go:336> smartstore: creating ro connection pool for shard module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc smartstore.shard_key=default
09:30:22 INFO <log/with.go:31> Starting HTTP listener module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc http.address=:8000
09:30:22 INFO <gomaxprocs/gomaxprocs.go:35> maxprocs.Set module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc message="maxprocs: Updating GOMAXPROCS=5: determined from CPU quota"
09:30:22 INFO <grpcserver/grpcserver.go:171> Serving tombstone v1 grpc server Port=5005 GracefulStopTimeout=15s DefaultUnaryRequestTimeout=10s ConnectionTimeout=10s KeepAliveMinTime=30s MaxConnectionIdle=2m0s MaxConnectionAge=5m0s MaxConnectionAgeGrace=10s KeepAliveTime=5s KeepAliveTimeout=30s MaxRecvMsgSize=10000 MaxSendMsgSize=10000
09:30:22 INFO <servicetoken/context.go:145> fetching service account module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc k8s_namespace=tombstone--bento1a service_account=tombstone-svc
09:30:22 INFO <servicetoken/fetcher.go:77> k8s service account token fetch module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc k8s_namespace=tombstone--bento1a service_account=tombstone-svc
09:30:22 INFO <lifecycle/server.go:124> Lifecycle hooks initialized module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc attempts=1
09:30:22 INFO <tombstone/rpc.go:109> Serving GRPC Service on :5000 module=github.com/getoutreach/gobox modulever=v1.110.4-0.20251019030948-893ef604dfcc
```


pkg/log/log.go
```diff
@@ -139,6 +211,11 @@ func Error(ctx context.Context, message string, m ...Marshaler) {
 
 // Fatal emits a log at FATAL level and exits.  This is for catastrophic unrecoverable errors.
 func Fatal(ctx context.Context, message string, m ...Marshaler) {
+	if ShouldUseSlog() {
+		slogIt(ctx, slog.LevelError, message, m)
+		os.Exit(1)
```
malept, line 228: **issue (non-blocking):** `Fatal` logs at `slog.LevelError`, making it indistinguishable from a regular `Error` log in the output.

The original non-slog path writes `"level": "FATAL"` along with structured `error.kind`, `error.message`, and `error.stack` fields via `generateFatalFields()`. With slog enabled, all of this is lost — consumers relying on `FATAL` level in log aggregation or alerting will silently miss these.

Consider using a custom slog level above `ERROR` (e.g., `slog.Level(12)` with a `"FATAL"` label) and/or adding the fatal fields to the slog record.

pkg/log/log.go
```diff
@@ -242,3 +319,101 @@ func generateFatalFields(entry F) {
...
+	// maps are unsorted, so we create a slice we can sort
+	type keyValue struct {
+		key   string
+		value any
+	}
+
+	var kvs []keyValue
+
+	logf.Marshal("", arg, func(key string, value any) {
+		kvs = append(kvs, keyValue{key: key, value: value})
```
malept, line 348: **suggestion (non-blocking):** This `sort.Slice` is redundant — the `[]slog.Attr` result is sorted again by key at the end of the function (line ~414). Since the `for` loop iterates `kvs` in sorted order and appends to `res` in that same order, the second sort alone is sufficient.

Remove this first sort to avoid the extra work.

pkg/log/log.go
```diff
@@ -242,3 +319,101 @@ func generateFatalFields(entry F) {
...
+			res = append(res, slog.Uint64(kv.key, uint64(v)))
+		case uint16:
+			res = append(res, slog.Uint64(kv.key, uint64(v)))
+		case uint32:
+			res = append(res, slog.Uint64(kv.key, uint64(v)))
+		case uint64:
+			res = append(res, slog.Uint64(kv.key, v))
+		case float32:
+			f64 := float64(v)
+			// Handle special float values that cause JSON encoding issues
```
malept, line 384: **issue (non-blocking):** The type switch handles `uint8`, `uint16`, `uint32`, `uint64`, but not `uint`. A value of type `uint` (e.g., `log.F{"count": uint(42)}`) will fall through to the `default` case and be string-formatted via `fmt.Sprintf`, producing `"42"` instead of a numeric `42`.

This is inconsistent with how `int` is handled (which has its own case). Consider adding:
```go
case uint:
    res = append(res, slog.Uint64(kv.key, uint64(v)))
```

pkg/log/log.go
```diff
@@ -55,9 +62,40 @@ var (
 	stdOut     io.Writer = &syncWriter{w: os.Stdout}
 	errOut     io.Writer = &syncWriter{w: os.Stderr}
 
+	// once to ensure we only initialize the slog.Logger once.
+	once          = sync.Once{}
+	slogLock      = sync.Mutex{}
+	_, shouldSlog = os.LookupEnv("GOBOX_AS_SLOG_FACADE")
```
malept, line 69: **thought:** The `sync.Once` guards first-use initialization of `log`, but `SetShouldUseSlog()` and `SetOutput()` both write to `log` directly, bypassing `once`. This is functionally correct (the `once` handles the lazy-init path, and the setters handle explicit re-init), but the interaction is non-obvious.

A comment here explaining this design would help future maintainers — e.g., that `once` only protects the first implicit initialization, while `SetShouldUseSlog`/`SetOutput` handle explicit re-initialization independently.

pkg/log/slog_test.go
```diff
@@ -0,0 +1,515 @@
...
+
+	// Return cleanup function
+	return func() {
+		// Restore original log output first
+		log.SetOutput(originalOutput)
+
+		// Then restore environment
+		var err error
+
+		log.SetShouldUseSlog(original)
```
malept, line 50: **nitpick:** `err` is declared but never assigned — it's always `nil`, so `assert.Check(t, err)` is a no-op. This looks like leftover code from when the cleanup may have called `os.Setenv`/`os.Unsetenv`. Remove both lines.

pkg/log/log.go
```diff
@@ -109,28 +153,56 @@ func Write(s string) {
...
+// slogIt produces a slog structured log at the appropriate level.
+func slogIt(ctx context.Context, lvl slog.Level, message string, m []Marshaler) {
+	once.Do(setupSlog)
+	var pcs [1]uintptr
+
+	runtime.Callers(3, pcs[:]) // skip [Callers, slogIt, gobox log function]
+
+	r := slog.NewRecord(time.Now(), lvl, message, pcs[0])
+	r.AddAttrs(slogAttrs(m)...)
+	_ = log.Handler().Handle(ctx, r) //nolint: errcheck //Why: mimic stdlib which skips handling this error
```
malept: **note:** `slogIt` reads the package-level `log` variable (line 165: `log.Handler().Handle(...)`) without holding `slogLock`, while `setupSlog()` and `SetShouldUseSlog()` write to `log` under the lock. After the first `once.Do` fires, subsequent calls skip `once.Do` entirely and read `log` unsynchronized — racing with any concurrent call to the setters.

In practice, the setters are typically called at startup so this is unlikely to matter in production, but it will trigger Go's race detector in tests that call `SetShouldUseSlog` concurrently with logging.

pkg/ometrics/go.mod
malept, line 1: **question (non-blocking):** `prometheus/procfs` went from `v0.20.1` to `v0.19.2` and `go.yaml.in/yaml/v2` from `v2.4.4` to `v2.4.3` — both are downgrades. Was this intentional (e.g., aligning with gobox `v1.111.4`'s dependency tree)?

pkg/log/log.go
malept: From the [docs](https://pkg.go.dev/log/slog#Level):

> OpenTelemetry also has the names TRACE and FATAL, which slog does not. But those OpenTelemetry levels can still be represented as slog Levels by using the appropriate integers.

We should probably set that to the OpenTelemetry FATAL value.

pkg/log/log.go
```diff
@@ -109,28 +153,56 @@ func Write(s string) {
 //	log.Error(ctx, "some failure", events.Err(err))
 type F = logf.F
 
+// slogIt produces a slog structured log at the appropriate level.
```
malept, line 157: thought: there should probably be a bit more description of what this does, for example the part about what I think is skipping some items from a potential backtrace.

malept (commented):
All but two comments are AI-generated, since the PR was created before we started doing AI code reviews. It should be obvious which are which.

pkg/log/log.go
AndrewWinterman: done

pkg/log/log.go
AndrewWinterman: done

pkg/log/log.go
AndrewWinterman: done, although I think the level will be `"level": "ERROR+13"` or something.

Do we want to correct that in this PR? I think it would require some modifications in olog and I'd kinda like that to happen in its own PR. Remember adopting this is gated behind ALSO setting an environment variable

pkg/log/log.go
AndrewWinterman: thanks

pkg/log/log.go
AndrewWinterman: done

pkg/log/slog_test.go
AndrewWinterman: fixed

pkg/log/log.go
```diff
@@ -242,3 +331,99 @@ func generateFatalFields(entry F) {
...
+	}
+
+	var kvs []keyValue
+
+	logf.Marshal("", arg, func(key string, value any) {
+		kvs = append(kvs, keyValue{key: key, value: value})
+	})
+
+	// Sort by key to ensure consistent ordering
+	sort.Slice(kvs, func(i, j int) bool {
```
evadnoob, line 352: I imagine this maintains some existing functionality, but is this cost worth the operation?

evadnoob (approved):

pkg/log/log.go
AndrewWinterman: basically both the logfmt and json handler (somehow) seem to respect order. Without this every invocation of even the same log line has a different order, which confuses human eyeballs.

There's likely no more than 10 fields, and almost certainly less than 100, so this should still be very snappy

richardlarocque (approved):
LGTM.

PR #742, opened 2025-11-14
fix(logger): add With method to logger

asms:
We can create loggers with logging context, but we can not append to that logging context. Add a `With` method to loggers to allow creating child loggers with both the parent logging context and additional context.

See: https://github.com/getoutreach/database-provisioning-operator/pull/352#discussion_r2529042776


AndrewWinterman (commented):
Can we validate this will work with #740 ? Also can you demo this working in another service?

PR #214, opened 2022-08-24
fix: use noop metrics provider for otel

artemchernyak-outreach:
## What this PR does / why we need it

Uses the noop metrics provider to prevent memory leaks.

## Jira ID

[DT-2744]

## Notes for your reviewers

[DT-2744]: https://outreach-io.atlassian.net/browse/DT-2744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #835, opened 2026-08-17
fix(trace): stop InitTracer installing a noop global MeterProvider

AndrewWinterman:
## What
Delete unconditional `otel.SetMeterProvider(noop)` in `initTracer` (+ unused import). 4 lines removed.

## Why
`trace.InitTracer` clobber real meter provider if service install one (order-dependent). Root cause of polaroid silent OTLP metric drop + FND-1013. No single owner of global.

## Safety analysis
- Noop line exist to guard otel-go global-delegate unbounded memory leak ([#5753](https://github.com/open-telemetry/opentelemetry-go/issues/5753)). Leak fixed upstream v1.30.0 ([#5754](https://github.com/open-telemetry/opentelemetry-go/pull/5754), follow-ups [#5780](https://github.com/open-telemetry/opentelemetry-go/pull/5780), [#5828](https://github.com/open-telemetry/opentelemetry-go/pull/5828)). gobox pin otel v1.45.0; Go MVS force all importers >= v1.45.0. Guard obsolete.
- Global delegate now dedupe identical instruments: memory bounded by unique instrument count. Measurements before real provider drop, never buffer — same as noop.
- `pkg/trace/logfile.go` noop-set untouched: dev/logfile/tracetest path behavior identical.
- Tests: `go test -tags or_test ./pkg/trace` — 11 fail with change, identical 11 fail on clean main (all `app.version "testing"` vs `"(devel)"` ldflags env artifact, unrelated). No regression.
- Build clean after removing unused `metric/noop` import.

## Behavior change (intended)
- Service that install real meter provider: no longer clobbered by `InitTracer`, any call order. Polaroid delete workaround after this.
- Service that never set provider: global stay delegating instead of noop. If real provider installed later, previously-dead metrics start exporting. Correct behavior, but silent metrics may appear after redeploy (dashboards, ingest volume) — heads up.

## Refs
FND-1013, FND-955, gobox #214 (introduced the guard)

### DIFF
--- a/pkg/trace/otel.go
+++ b/pkg/trace/otel.go
@@ -22,7 +22,6 @@ import (
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
-	"go.opentelemetry.io/otel/metric/noop"
 	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/sdk/resource"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
@@ -89,9 +88,6 @@ func (t *otelTracer) registerSpanProcessor(s sdktrace.SpanProcessor) {
 }
 
 func (t *otelTracer) initTracer(ctx context.Context, serviceName string) error {
-	mp := noop.NewMeterProvider()
-	otel.SetMeterProvider(mp)
-
 	var client otlptrace.Client
 
 	// We want to default to initialize and send traces through the OpenTelemetry collectors.


richardlarocque (commented):
For services that opt in to a specific provider, this seems obviously right.

For those that go back to the default ("defer provider"?) I'm a bit more worried.

On a hunch I followed the thread of `git blame` back to https://outreach-io.atlassian.net/browse/DT-2744 and https://app.slack.com/client/T02UUSZ58/CM3TW594L?skip_today=1.

It seems this noop thing was added in response to memory leaks observed in production.  If we make this change, will we see metrics buffered in-memory indefinitely, and more memory leaks?

AndrewWinterman:
@richardlarocque I had the same concern, so I asked claude about it.

It found a series of commits between when the no-op provider was set and the
current gobox version that resolve the otel memory leak. In short we will still
allocate some memory for unique otel metric names, but that *should* be a fixed
and small set of objects unless something weird is happening (e.g. org guid in MetricName)

Should we gate this behind a feature flag for extra safety?

Claude's somewhat claudish response included below:

>  Short answer: no, that was fixed in otel-go v1.30.0 (Sept 2024), and gobox is
>  on v1.45.0 — well past it.

>  Details of what the behavior was and is:

>  - The old leak: the global delegating MeterProvider (what otel.Meter() returns
>    before any SetMeterProvider) retained every instrument-creation call so it
>    could re-create them on a real provider later. Repeatedly creating the same
>    instrument — e.g. otelhttp.NewTransport() per request, or any
>    meter.Int64Counter(name) in a hot path — grew memory without bound forever
>    if no provider was ever set. Tracked as #5753 and contrib #5190 ("Memory
>    leak caused by otelhttp when no Meter Provider is configured").
>  - The fix: PR #5754 (milestone v1.30.0) makes the global delegate cache/dedupe
>    identical instruments, so repeated creation of the same instrument no longer
>    accumulates. Follow-up races it introduced were fixed in #5780 and #5828.
>    (An earlier partial fix, #4820, cached instruments in the SDK meter but
>    didn't cover the global delegate.)
>  - What's still retained: distinct instruments and observable-callback
>    registrations are still held indefinitely awaiting delegation — bounded by
>    the number of unique instruments, which is fine for any sane program (unless
>    someone puts unbounded cardinality in instrument names). Measurements
>    recorded before delegation are dropped, not buffered — they were never the
>    leak.

>  Implication for the plan: gobox's defensive noop.NewMeterProvider() in
>  pkg/trace/logfile.go:37 — the line PR #214 added to stop exactly this leak —
>  is guarding against an upstream bug that no longer exists at v1.45.0.

richardlarocque (approved):

kaldorn (approved):

PR #859, opened 2026-09-03
feat(orerr): respect gRPC status and context errors in classification

AndrewWinterman:
## What
`orerr.ExtractErrorStatusCode` / `ExtractErrorStatusCategory` previously only recognized `orerr.StatusCodeWrapper` and defaulted everything else to `InternalServerError` / `CategoryServerError`. This means `trace.SetCallError` classified errors carrying a gRPC status (e.g. `status.Errorf(codes.NotFound, ...)`) or `context.Canceled`/`DeadlineExceeded` as server errors in traces/metrics, even though the gRPC wire code returned to clients was correct.

## Changes
- Fallback chain in `ExtractErrorStatusCode`: `StatusCodeWrapper` (unchanged precedence) → `context.Canceled` → `statuscodes.Cancelled` → `context.DeadlineExceeded` → `statuscodes.DeadlineExceeded` → any `GRPCStatus()` error in the chain (via `errors.As`), mapped to the matching `statuscodes` value (inverse of the services/grpcx mapping, plus `AlreadyExists`→`Conflict`) → `InternalServerError` default.
- `ExtractErrorStatusCategory` now derives from the resolved code via `code.Category()`.
- Table tests for gRPC codes, context errors, wrapper precedence, and plain-error fallback.

## Test plan
- [x] `go build ./...`
- [x] `go test -tags or_test ./pkg/orerr/...`
- Note: `pkg/trace` tests fail identically on `origin/main` when run without `make test` ldflags (`modulever` mismatch) — pre-existing, unrelated.

## References
- `interface{ GRPCStatus() *status.Status }` is grpc-go's official error-integration contract, the same one `status.FromError` uses (including `errors.As` chain unwrapping since grpc-go v1.58): [pkg.go.dev/google.golang.org/grpc/status#FromError](https://pkg.go.dev/google.golang.org/grpc/status#FromError), [grpc-go rpc-errors.md](https://github.com/grpc/grpc-go/blob/master/Documentation/rpc-errors.md)
- Service-side counterpart: getoutreach/polaroid#590

### DIFF
--- a/pkg/orerr/statuscodewrap.go
+++ b/pkg/orerr/statuscodewrap.go
@@ -114,6 +114,8 @@ func grpcCodeToStatusCode(code codes.Code) statuscodes.StatusCode {
 		return statuscodes.Cancelled
 	case codes.AlreadyExists:
 		return statuscodes.Conflict
+	case codes.Unknown, codes.FailedPrecondition, codes.Aborted, codes.OutOfRange, codes.DataLoss:
+		return statuscodes.InternalServerError
 	default:
 		return statuscodes.InternalServerError
 	}
--- a/pkg/orerr/statuscodewrap.go
+++ b/pkg/orerr/statuscodewrap.go
@@ -5,9 +5,12 @@
 package orerr
 
 import (
+	"context"
 	"errors"
 
 	"github.com/getoutreach/gobox/pkg/statuscodes"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 type StatusCodeWrapper struct {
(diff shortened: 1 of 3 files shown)


pkg/orerr/statuscodewrap.go
```diff
@@ -62,13 +65,58 @@ func ExtractErrorStatusCode(err error) statuscodes.StatusCode {
...
+		return statuscodes.NotImplemented
+	case codes.Unavailable:
+		return statuscodes.Unavailable
+	case codes.DeadlineExceeded:
+		return statuscodes.DeadlineExceeded
+	case codes.Canceled:
+		return statuscodes.Cancelled
+	case codes.AlreadyExists:
+		return statuscodes.Conflict
+	case codes.Unknown, codes.FailedPrecondition, codes.Aborted, codes.OutOfRange, codes.DataLoss:
```
AndrewWinterman: Neither FailedPrecondition, OutofRange, or Aborted are server errors.

PR #866, opened 2026-09-11
fix: allow debug level logs for CLI tools

PeSme:

CLI tools like [smartstorectl](https://github.com/getoutreach/smartstorectl/blob/main/internal/tunnel/tunnel.go#L43) use TextHandler when executed in TTY. They accept `--loglevel debug` but do not honor it because there is no way how to change the log level globally for already created `charmbracelet/log` logger instances. And they are usually created in the init phase before the `main()` code executes and arguments are processed. 

The linked [issue](https://github.com/charmbracelet/log/issues/98) was closed as a won't fix and if I understood correctly this was suggested as a solution. It's ugly but it works well. 

Of course, it will need a code change in every tool using it. After parsing `--loglevel` for `logrus` the script will need to update `olog` as well for example [here](https://github.com/getoutreach/smartstorectl/blob/main/cmd/smartstorectl/psql/commands.go#L73) like this:

```
				var level slog.Level
				if err := level.UnmarshalText([]byte(lvl)); err != nil {
					return err
				}
				olog.SetGlobalLevel(level)
				return nil
```

## Jira ID

[XX-XX]

## Notes for your reviewers

`charmlog.Logger` implements `slog.Handler`, but its `Enabled`/`Handle` methods only consult a level set once via `charmlog.Options.Level` or `(*charmlog.Logger).SetLevel` — it has no concept of a `slog.Leveler` that gets re-evaluated on every log call. So a bare `*charmlog.Logger` can't pick up level changes made after it was created (e.g. via `SetGlobalLevel`), which is exactly the problem for `olog`'s package-level `log` singletons that are created well before a program parses its flags.

`charmLevelHandler` (in `pkg/olog/default_handler.go`) works around this by wrapping the `*charmlog.Logger` and re-syncing its level from our own `slog.Leveler` every time `Enabled` is called — which `slog` always does immediately before `Handle` for a given record — so the wrapped logger's level is always brought up to date before it gets a chance to reject or accept the record.

### DIFF
--- a/pkg/olog/default_handler.go
+++ b/pkg/olog/default_handler.go
@@ -6,6 +6,7 @@
 package olog
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"log/slog"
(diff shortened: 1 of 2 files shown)


pkg/olog/olog_test.go
```diff
@@ -145,3 +146,47 @@ func TestOutputLog(t *testing.T) {
...
+// TestTextHandlerRespectsDynamicLevel ensures that loggers using the
+// TextHandler (backed by charm.land/log/v2) respect log-level changes
+// made via SetGlobalLevel *after* the logger has already been created.
+//
+// This is a regression test: charmlog.Logger does not implement
+// slog.Leveler, it only supports a level set once at creation (or via
+// SetLevel). createHandler wraps it in charmLevelHandler specifically
+// to keep it in sync with the leveler dynamically -- without that
+// wrapper, this test would fail because the DEBUG line logged after
+// SetGlobalLevel would not appear.
```
ekalinichev:
```suggestion
```

pkg/olog/default_handler.go
```diff
@@ -179,3 +172,63 @@ func replaceKey(oldKey, newKey string) func([]string, slog.Attr) slog.Attr {
...
+// on its own can't pick up changes made after it was created (e.g. via
+// SetGlobalLevel), which is a problem for `log` package-level
+// singletons that are created well before a program has parsed its
+// configuration/flags and decided what level it wants.
+//
+// charmLevelHandler works around this by re-syncing the wrapped
+// logger's level from `leveler` every time Enabled is called (which
+// slog always does immediately before Handle for a given record), so
+// the charm logger's level is always brought up to date before it
+// gets a chance to reject or accept the record.
```
ekalinichev: I don't mind the fix but let's shorten this _significantly_ and move to PR description if it's not already there. 

```suggestion
// charmLevelHandler is a slog.Handler that wraps a *charmlog.Logger in
// order to make it respect a slog.Leveler dynamically.
```

ekalinichev (changes requested):
Fix is nice, just need comments to be less verbose and it's good to go

pkg/olog/default_handler.go
PeSme: @ekalinichev , done

ekalinichev (approved):

HyeonJi-Kang (approved):

PR #871, opened 2026-09-22
feat: added otelmetrics library for OTLP metrics

alexandrkara-outreach:

The library includes both push metrics and pull metrics (scraped externally). The intention is to use them from stencil'ed main() and register the push Service if enabled in service.yaml.

Then, if enabled, instrument standard gRPC and Kafka libraries to emit standardized metrics.

## Jira ID

[SEARCH-2773]

## Notes for your reviewers

TESTED=make test

[SEARCH-2773]: https://outreach-io.atlassian.net/browse/SEARCH-2773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/otelmetrics/otelmetrics.go
```diff
@@ -0,0 +1,259 @@
...
+	return singletonService.Load()
+}
+
+// Run implements async.Runner - initializes OTel and blocks until the context is cancelled.
+// A failure to activate the registered instruments is considered fatal.
+func (s *Service) Run(ctx context.Context) error {
+	if !s.cfg.Enabled {
+		log.Info(ctx, "OTLP push metrics disabled, skipping initialization")
+		<-ctx.Done()
+		return nil
```
ekalinichev: **Critical:** When `Config.Enabled` is `false`, `Run()` blocks on `<-ctx.Done()` and returns without ever calling `close(s.ready)` (that only happens later, on the enabled path, after `activate()` succeeds at line 146). But `Ready()`'s doc comment promises the channel closes "once instrument activation has completed successfully and metrics recording is live" — it doesn't say "or never, if disabled."

Any caller that does `<-otelmetrics.SingletonService().Ready()` before proceeding (a natural pattern for "wait until metrics plumbing is live") will hang forever whenever OTLP push metrics are disabled in `service.yaml` — which is a completely normal deployment state, not an edge case.

Suggest closing `s.ready` on the disabled branch too (or renaming/documenting `Ready()` so callers know a disabled service never becomes ready and must not block on it).

pkg/otelmetrics/otelmetrics.go
```diff
@@ -0,0 +1,259 @@
...
+	// context because the handler may fire after Run's context is cancelled
+	// during shutdown.
+	otel.SetErrorHandler(otel.ErrorHandlerFunc(func(err error) {
+		log.Error(ctx, "OTel metrics error", events.NewErrorInfo(err))
+	}))
+
+	if err := s.initialize(ctx); err != nil {
+		// Log but do not fail startup - metrics are non-critical.
+		log.Error(ctx, "Failed to initialize OTLP push metrics, continuing without", events.NewErrorInfo(err))
+		return err
```
ekalinichev: **Critical:** This comment and the code disagree: "Log but do not fail startup - metrics are non-critical" is immediately followed by `return err`.

I dug into how this plays out via `pkg/async`:
- If this `Service` is driven through `async.Run`/`Tasks.Run`, a non-nil `Run()` error is just logged (`async.go:75-84`) — not fatal to the process. But because `Run` returns immediately here instead of blocking/retrying, the metrics service goroutine simply ends. Nothing calls `Run` again, so OTLP push metrics are silently dead for the rest of the process lifetime, with a single log line and no retry. `Ready()` never closes either (see the other comment), so anything downstream waiting on it hangs. That's arguably worse than "continuing without metrics" as the comment claims - it's "permanently and silently broken."
- If instead this `Service` is put into an `async.RunGroup` alongside other runners (e.g. the gRPC server), `errgroup` semantics mean this error cancels the shared context for every other runner in that group (`async.go:142-158`) — i.e. a transient OTLP endpoint hiccup at startup could take down unrelated runners sharing the group, which is the opposite of "non-critical."

Either way, I don't think this should short-circuit `Run()` on a single failed attempt - have `Run()` keep running (block on `ctx.Done()`) after logging so the process behavior actually matches "non-critical, continue without it." (See root comment re: retry.)

pkg/otelmetrics/otelmetrics.go
```diff
@@ -0,0 +1,259 @@
...
+		sdkmetric.WithReader(
+			sdkmetric.NewPeriodicReader(exporter,
+				sdkmetric.WithInterval(s.cfg.PushInterval),
+			),
+		),
+	)
+
+	// Register as the global meter provider so any package using otel.Meter
+	// picks it up.
+	otel.SetMeterProvider(s.meterProvider)
```
ekalinichev: **Critical:** This unconditionally overwrites the process-global OTel meter provider via `otel.SetMeterProvider`. `pkg/ometrics/ometrics.go`'s `InitializeMeterProvider` (specifically its `ExporterTypeCollector` case) already does the exact same OTLP-gRPC push setup and also calls `otel.SetMeterProvider` on the same global.

If any service (or a library dependency) ends up calling both `ometrics.InitializeMeterProvider(ctx, ometrics.ExporterTypeCollector, ...)` and `otelmetrics.NewService(...).Run(ctx)`, whichever finishes last silently wins — instruments already created against the first provider keep reporting to a now-orphaned provider, with zero error or warning. Given `pkg/ometrics` already exists in this repo and does largely the same thing (just with an older semconv pin, v1.21.0 vs this package's v1.26.0), it'd be worth either building on `pkg/ometrics` directly or adding an explicit guard against double-setting the global provider.

pkg/otelmetrics/registration.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// trySend enqueues item onto the bounded emit queue without blocking and returns whether it was successfully enqueued.
+func (m *MetricsOutput) trySend(item emitWorkItem) bool {
+	m.m.Lock()
+	ch := m.emitChannel
+	m.m.Unlock()
+
+	// If the emit channel is not yet active, the item cannot be enqueued.
+	if ch == nil {
```
ekalinichev: **Medium:** `trySend` returns `false` both when the queue is genuinely full *and* when `ch == nil` (service disabled, or not yet activated). Every `Add()`/`Record()` call site funnels both cases into the same `reportFullEmitBuffer` → `log.Warn(ctx, "Full emit buffer for OTLP metrics", ...)`.

Any service that runs with metrics disabled (or during the startup window before `activate()` runs) will log a misleading "buffer full" warning about once a minute forever, even though nothing is full — the queue was simply never created. Worth distinguishing "not active" from "full" so on-call doesn't chase a capacity problem that doesn't exist.

pkg/otelmetrics/otelmetrics.go
```diff
@@ -0,0 +1,259 @@
...
+func NewService(cfg *Config, serviceName string, opts ...Option) *Service {
+	service := &Service{
+		cfg:   cfg.WithDefaults(serviceName),
+		ready: make(chan struct{}),
+	}
+	for _, opt := range opts {
+		opt(service)
+	}
+
+	if singletonService.Load() != nil {
```
ekalinichev: **Medium:** This is a non-atomic check-then-act: `singletonService.Load() != nil` is read, then `Store()` happens separately. Two concurrent `NewService` calls can both pass the nil check before either stores, so the second call silently overwrites the first singleton instead of hitting the panic this code is clearly trying to enforce.

Suggest `if !singletonService.CompareAndSwap(nil, service) { panic(...) }` instead.

pkg/otelmetrics/otelmetrics.go
```diff
@@ -0,0 +1,259 @@
...
+	if err := singletonMetricsOutput.activate(meter, emitChannel); err != nil {
+		return errors.Wrap(err, "failed to activate OTLP metric instruments")
+	}
+
+	log.Info(ctx, "OTLP metric instruments activated")
+	close(s.ready)
+
+	for i := 0; i < max(s.cfg.PushWorkerCount, 1); i++ {
+		go s.worker(ctx, emitChannel)
+	}
```
ekalinichev: **Medium:** The worker goroutines spawned here aren't joined anywhere. `Run()` returns as soon as `ctx` is cancelled, and the lifecycle then calls `Close()` (which calls `meterProvider.Shutdown()` with a 10s timeout) with no guarantee the workers have actually stopped calling `item.emit(ctx)` against instruments owned by that same provider. Consider a `sync.WaitGroup` around the worker goroutines that `Close()` (or the end of `Run()`) waits on before/while shutting down the provider, to avoid dropping metrics recorded right at shutdown.

pkg/otelmetrics/metrics.go
```diff
@@ -0,0 +1,49 @@
...
+		if lastReported != nil && now.Sub(*lastReported) <= time.Minute {
+			return
+		}
+
+		if swapped := lastReportedError.CompareAndSwap(lastReported, &now); swapped {
+			break
+		}
+	}
+
+	log.Warn(ctx, "Full emit buffer for OTLP metrics", log.F{"metric_name": name})
```
ekalinichev: **Medium:** This loop's own rate-limit guarantee ("at most once per minute") can be defeated under contention: if all 5 `CompareAndSwap` attempts lose the race (plausible exactly when many goroutines are hitting a genuinely full buffer at once - the scenario this exists to protect against), the loop falls through to an unconditional `log.Warn` regardless of whether the swap ever succeeded. A single unconditional `CompareAndSwap` attempt (return without logging if it fails, since another goroutine just logged or is about to) would be simpler and wouldn't have a path that silently breaks its own contract.

pkg/otelmetrics/config.go
```diff
@@ -0,0 +1,76 @@
...
+// WithDefaults returns a copy of the config with default values applied.
+func (c Config) WithDefaults(serviceName string) Config {
+	cfg := c
+	if cfg.ServiceName == "" {
+		cfg.ServiceName = serviceName
+	}
+	if cfg.PushInterval <= 0 {
+		cfg.PushInterval = 30 * time.Second
+	}
+	if cfg.ExportTimeout == 0 {
```
ekalinichev: **Medium:** This checks `== 0` while `PushInterval`, `PushWorkerCount`, and `EmitBufferSize` all use `<= 0`. A negative `ExportTimeout` (bad YAML/env parsing, copy-paste error) skips the default and flows straight into `otlpmetricgrpc.WithTimeout(negative)` in `initialize()`. Suggest matching the other three fields with `<= 0`.

pkg/otelmetrics/metrics_type_float64_counter.go
```diff
@@ -0,0 +1,80 @@
...
+
+package otelmetrics
+
+import (
+	"context"
+
+	"go.opentelemetry.io/otel/metric"
+)
+
+// Float64CounterMetric represents a float64 histogram metric.
```
ekalinichev: **Medium:** Copy-paste error: "`Float64CounterMetric represents a float64 histogram metric.`" (should say "counter", not "histogram" - looks copied from the histogram file's comment).

More broadly, this is one symptom of a deeper reuse issue: the 8 metric-type files (`int64`/`float64` × `counter`/`gauge`/`histogram`/`up_down`) hand-duplicate the same `Add`/`Record` + `trySend` + `Register` pattern, differing only in the concrete OTel instrument type. That's ~500 lines of copy-pasted logic where a single generic implementation (Go 1.25 generics are already used elsewhere in gobox, e.g. `pkg/set`, `pkg/pointer`) could collapse this to one implementation parameterized over the instrument type, and this doc-comment slip is direct evidence the copy-paste discipline has already lapsed once. Worth a follow-up if not in scope for this PR.

pkg/otelmetrics/registration.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil
+	}
+
+	m.instrumentRegistrars[metricName] = registrationFn
+
+	return nil
+}
+
+// trySend enqueues item onto the bounded emit queue without blocking and returns whether it was successfully enqueued.
+func (m *MetricsOutput) trySend(item emitWorkItem) bool {
```
ekalinichev: **Medium:** `trySend` takes the process-wide `MetricsOutput` mutex on every single `Add()`/`Record()` call across the whole library just to read `m.emitChannel`, a pointer that's written exactly once (in `activate()`). Every `Add`/`Record` call site (in the 8 metric-type files) also heap-allocates a closure to defer a call into the OTel SDK that's already safe for concurrent synchronous use. Since this library is meant to sit on the gRPC/Kafka hot path per the PR description, that's a global lock plus an allocation on every recorded metric with no correctness benefit. Consider storing `emitChannel` in an `atomic.Pointer` (the same pattern already used for `singletonService`) to drop the lock from the hot path, and/or recording synchronously given the underlying OTel call is already non-blocking and thread-safe.

pkg/otelmetrics/registration_test.go
```diff
@@ -0,0 +1,195 @@
...
+//
+// ObservableGroup defers each instrument's creation into its own pending
+// slice and only touches the singleton once, from Prepare - so the "meter
+// still nil at registration time" path (what actually happens in
+// production: instruments register from init() before Service.Run ever
+// creates a meter, and MetricsOutput.activate() then runs every deferred
+// registrar in Go map iteration order) is just as safe: Prepare's single map
+// entry runs pending (in order) before RegisterCallback, regardless of where
+// in that random iteration it lands. TestKafkaWriterInstruments in
+// kafkawriter_test.go exercises that real, deferred path end-to-end.
```
ekalinichev: **Low:** This comment claims `kafkawriter_test.go` exercises the "meter still nil at registration time" path end-to-end, but neither `kafkawriter.go` nor `kafkawriter_test.go` exist anywhere in this repo (checked `git log --all` / `git grep`). Looks like a forward-reference to a planned downstream consumer that hasn't landed yet - worth rewording so it doesn't read as an existing, verified test dependency.

pkg/otelmetrics/registration.go
```diff
@@ -0,0 +1,138 @@
...
+// ObservableGroup collects a set of observable instruments that are reported together from a single OTel callback.
+type ObservableGroup struct {
+	groupName   string
+	pending     []func(meter metric.Meter) error
+	observables []metric.Observable
+}
+
+// Prepare finalizes the group once the meter is available and registers collectFn to be called at every collection time.
+func (g *ObservableGroup) Prepare(groupName string, collectFn func(context.Context, Observer) error) error {
+	g.groupName = groupName
```
ekalinichev: **Low:** `g.groupName` is written here but never read anywhere else - `Prepare` uses the local `groupName` parameter directly (e.g. line 137's `MetricName(groupName)`). Same pattern for `observableMetricBase.name` in `metrics.go:23` (set in every `RegisterXObservableY` constructor, e.g. `metrics_type_float64_counter.go:66`, but never read - the observable `Record` methods call `o.ObserveFloat64/ObserveInt64` directly). Both look like dead write-only bookkeeping that could be dropped.

ekalinichev (changes requested):
Reviewed `pkg/otelmetrics`. Overall the design (deferred registration via `init()`, a bounded async emit queue, delta temporality for the ClickHouse backend) is solid, but there are a few correctness issues around the disabled/failure paths that I think should be fixed before this lands, plus some medium/low-priority cleanup. Details inline.

One general note: a couple of the failure paths below (e.g. the OTLP init failure in `Run()`) look like they'd benefit from retry/backoff rather than a single-shot failure - consider https://github.com/getoutreach/services/tree/main/pkg/retry for retry functionality if applicable.

pkg/otelmetrics/otelmetrics.go
alexandrkara-outreach: It's genuinely hard, if we signaled "Ready" but the service was disabled or initialization failed, we still wouldn't emit anything. I renamed it so that it's only used in tests (which is the current situation), there shouldn't be a need to know otherwise.

pkg/otelmetrics/otelmetrics.go
alexandrkara-outreach: The comment was wrong (from an earlier state where intialization was not fatal). I checked and it looks like it creates gRPC clients (which shouldn't perform I/O so shouldn't spuriously fail) and if it fails, we are likely going to be without metrics and retries wouldn't help.

We could either start or refuse starting, which is probably a better option - the pod shouldn't get into "ready" state so a broken deployment should be stopped.

pkg/otelmetrics/otelmetrics.go
alexandrkara-outreach: There are actually 3 packages that set this singleton: pkg/ometrics, pkg/trace and pkg/otelmetrics. They shouldn't really be used together but if they do, I think it could still work since this is the implementation:
```
func SetMeterProvider(mp metric.MeterProvider) {
	current := MeterProvider()
	if _, cOk := current.(*meterProvider); cOk {
		if _, mpOk := mp.(*meterProvider); mpOk && current == mp {
			// Do not assign the default delegating MeterProvider to delegate
			// to itself.
			Error(
				errors.New("no delegate configured in meter provider"),
				"Setting meter provider to its current value. No delegate will be configured",
			)
			return
		}
	}

	delegateMeterOnce.Do(func() {
		if def, ok := current.(*meterProvider); ok {
			def.setDelegate(mp)
		}
	})
	globalMeterProvider.Store(meterProviderHolder{mp: mp})
}
```

pkg/otelmetrics/registration.go
alexandrkara-outreach: Makes sense, changed to log a different message.

pkg/otelmetrics/otelmetrics.go
alexandrkara-outreach: Makes sense, changed to CompareAndSwap().

pkg/otelmetrics/otelmetrics.go
alexandrkara-outreach: Right, better to use sync.WaitGroup so that no metrics are hopefully lost.

pkg/otelmetrics/metrics.go
alexandrkara-outreach: Right, it's simpler that way.

pkg/otelmetrics/config.go
alexandrkara-outreach: Changed all to <= 0.

pkg/otelmetrics/metrics_type_float64_counter.go
alexandrkara-outreach: Good catch, changed the comment. On removing the copy-paste logic, it's not so simple because all the metrics differ not just in their type (int64, float64) but also the type of metric (some use Add(), some Record()) and also the underlying types are different (e.g. metric.Float64ObservableGauge or metric.Int64Counter) and also their arguments are different (e.g. metric.Int64HistogramOption or metric.Float64ObservableGaugeOption) so to bind all together, maybe the amount of boilerplate is not too bad.

pkg/otelmetrics/registration.go
alexandrkara-outreach: OK, makes sense, changing to atomic.Pointer.

pkg/otelmetrics/registration_test.go
alexandrkara-outreach: It's a reference to searchindexer, I removed it to not be confusing.

pkg/otelmetrics/registration.go
alexandrkara-outreach: OK, I planned to possibly log it if needed but true, it's not used at the moment, so removing.

